PR #2, opened 2019-08-08
Apply lifecycle rule for big objects on primary data bucket

brainstorm:
@reisingerf Not applied yet... review please? ;)

### DIFF
--- a/terraform/stacks/bootstrap/main.tf
+++ b/terraform/stacks/bootstrap/main.tf
@@ -287,11 +287,21 @@ resource "aws_s3_bucket" "primary_data" {
       days = 90
     }
 
-	expiration {
-		expired_object_delete_marker = true
-	}
+    expiration {
+      expired_object_delete_marker = true
+    }
 
-	abort_incomplete_multipart_upload_days = 7
+    abort_incomplete_multipart_upload_days = 7
+  }
+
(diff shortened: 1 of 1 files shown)


terraform/stacks/bootstrap/main.tf
```diff
@@ -287,11 +287,21 @@ resource "aws_s3_bucket" "primary_data" {
...
+      expired_object_delete_marker = true
+    }
+
+    abort_incomplete_multipart_upload_days = 7
+  }
 
-	abort_incomplete_multipart_upload_days = 7
+  lifecycle_rule {
+    id      = "Deep archive BAM files older than 90 days"
+    filter  = "*.bam"
```
reisingerf: Where did you get this notation from?
First I don't think the `lifecycle_rule` supports that `filter` attribute, and second AFAIK S3 does not support wildcard filters, only prefix filters.

brainstorm:
https://www.terraform.io/docs/providers/aws/r/s3_bucket.html

Skärmavbild 2019-08-09 kl  09 32 50

You are right, that refers to `rules` not `lifecycle_rules`, my bad. I guess it would have to be based on `prefix` or tags somehow then.

reisingerf:
Closing as current approach is not supported.

PR #14, opened 2020-02-23
Added samplesheet, created gridss formation example, cloudformation appropriate for onboarding user

alexiswl:
samplesheet logic for creating bat file and anaconda environment on a windows computer
created worker.yaml for gridss exploration as a useful example for adapting cloudformation templates,
created an onboarding yaml file similar to the dev yaml file with different credentials.


cloudformation/templates/gridss_worker_instance/worker.yaml
```diff
@@ -0,0 +1,221 @@
...
+            amazon-linux-extras install epel -y
+            yum install s3fs-fuse -y
+            # Sync gridss data from s3 bucket
+            aws s3 sync --quiet s3://umccr-misc-temp/gridss_hg19_refdata/hg19/ /mnt/xvdh/hg19_gridss_ref_data/
+            # Install container registry helper
+            yum install amazon-ecr-credential-helper -y
+            # Add configuration to docker config
+            su - "ec2-user" -c 'mkdir -p $HOME/.docker && echo "{ \"credsStore\" : \"ecr-login\" }" >> $HOME/.docker/config.json'
+            # Pull the gridss-purple-linx container image from the ECR
+            su - "ec2-user" -c 'docker pull 843407916570.dkr.ecr.ap-southeast-2.amazonaws.com/gridss-purple-linx:2.7.3'
```
brainstorm: @alexiswl Could you parametrise this out somehow? I almost missed it and can be fiddly to update if buried down in userdata :-S

cloudformation/templates/gridss_worker_instance/worker.yaml
```diff
@@ -0,0 +1,221 @@
...
+            mv /var/lib/docker /mnt/xvdh/docker
+            # Link mount to docker container
+            ln -s /mnt/xvdh/docker /var/lib/docker
+            # Add users to docker
+            usermod -a -G docker ssm-user
+            usermod -a -G docker ec2-user
+            service docker start
+            # Install s3fuse from amazon
+            amazon-linux-extras install epel -y
+            yum install s3fs-fuse -y
```
brainstorm: Why is fuse needed here? Normally with plain `aws cli` you get better performance, reliability (MD5 checksums per multipart) and stability. With S3FS you are renouncing to all that :/

cloudformation/templates/gridss_worker_instance/worker.yaml
```diff
@@ -0,0 +1,221 @@
...
+          - Effect: Allow
+            Principal:
+              Service:
+                - ec2.amazonaws.com
+            Action:
+              - 'sts:AssumeRole'
+      Path: /worker/
+      ManagedPolicyArns:
+        - 'arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess'
+        - 'arn:aws:iam::843407916570:policy/alexis_mnt_test_write_policy'
```
brainstorm: Is there a managed policy that could potentially substitute this custom policy?

cloudformation/templates/gridss_worker_instance/worker.yaml
alexiswl: As in the ECR container registry link and the docker image and tag names?

brainstorm: Yeah, sth like this?: https://github.com/umccr/umccrise/blob/master/buildspec.yml#L4

cloudformation/templates/gridss_worker_instance/worker.yaml
alexiswl: Yep, left over from a previous experiment set seeing if I could mount an S3, turns out I can without too much stress but later learning taught me that it wasn't common practise, hence there's not much documentation out there for it. I'll remove it from the bootstrap

cloudformation/templates/gridss_worker_instance/worker.yaml
alexiswl: I needed this initially to add the bwa indices to the ref data (and then re-sync via aws s3) to the bucket and this policy gave me write access to that. I don't think it needs to be there now, but if a user wants to save the data to any bucket they'll need to add a policy to the running instance. Do we have a 'dump' bucket we can code into a policy or are users expected to add in their own policies retrospectively once they know where they want their outputs to go?

brainstorm: We have a (per-app) reference data bucket on dev for BWA indices and a bunch of others (`umccr-refdata-dev` or similar)... ideally there should be as few custom policies as possible and those type of resources should be shared if they are heavyweight like refdata. You are right with output files, I can't think of any other way to define and restrict them on a case by case basis.

cloudformation/templates/gridss_worker_instance/worker.yaml
alexiswl: I've used the AWS::AccountId and AWS::Region envs, which I get might cause some problems if the container doesn't exist in that account / region but we can always just duplicate the container to multiple registries, it makes more sense to grab it from the registry from the corresponding account and fail if it doesn't contain the repo.

cloudformation/templates/gridss_worker_instance/worker.yaml
alexiswl: This was a specific hg19 reference as mentioned here https://github.com/hartwigmedical/gridss-purple-linx#reference-data but I'll note that bucket for future reference

PR #19, opened 2020-03-02
Gridss cdk example

alexiswl:
Created an example for running gridss-purple-lynx cdk ec2 instance through aws cdk.


alexiswl:
@brainstorm I placed the semi-global parameters in the cdk.json file and provided an example of changing them in the readme. Yet to determine how to change nested parameters. Might not be possible at this stage.

alexiswl:
Deprecated by #21

PR #20, opened 2020-03-03
add user ID

pdiakumis:


cdk/apps/slack/lambdas/iap/notify_slack.py
```diff
@@ -36,7 +36,7 @@ def getAwsAccountName(id):
 def getCreatorFromId(id):
     if id == 'c9688651-7872-3753-8146-ffa41c177aa1':
         return f"{id} (Vlad Saveliev)"
-    elif id == '567d89e4-de8b-3688-a733-d2a979eb510e':
+    elif id == 'bd68e368-7587-3e22-a30e-9a2e1714b7c1':
```
reisingerf: Can you keep the old ID (see other users), just in case some of the older messages/ghost runs keep popping up?

pdiakumis: Yep, sorry ;-)

PR #32, opened 2020-03-17
Added Spot Pricing to dev worker instance

alexiswl:
Additional changes:
userdata now installs IAP.
Updated to latest cdk (1.28.0)

PR #33, opened 2020-03-27
Added STACK_NAME key to context

alexiswl:
stack names can now be dynamic following up issue raised in https://github.com/umccr/infrastructure/pull/32


alexiswl:
I have no idea why 8/9 commits are commits that have already been merged in #32, maybe I needed to update my fork from the upstream prior to making the commit

alexiswl:
Force push doesn't seem to solve it. Bit weird, okay to merge though.

alexiswl:
Whoops

brainstorm:
So, practically the `STACK_NAME` is meant to help avoid deploy collisions between our users... would there be a way to parametrize it with `$USER` or the like somehow?

reisingerf:
You mean pick up `$USER` by default? Otherwise you could just use that on the command line...
That would go a little against the recommendations, but could be done...

I would not set a default in the `cdk.json` though and prepare the code for the case where that context variable is not present.

For example:
First try the context (that's in case someone is specifically setting it with `cdk synth -c STACK_NAME=foobar`, not from the `cdk.json`).
If no command line option was given, try to create a name based on the `$USER` env varriable (which would be the default, if not explicitly overwritten).
Lastly, we have to cater for the case that there's no `$USER` variable...
Something similar to the following:

```
stack_name = app.node.try_get_context("STACK_NAME")

if not stack_name and os.environ.get('USER'):
    stack_name = f"worder_{os.environ.get('USER')}"

if not stack_name:
    stack_name = "worker_default"
```

alexiswl:
In the last update, given that keys are also per-user, we should force that to be explicitly defined. I've changed the STACK_NAME and KEY_NAME keys to null as default. The `app.node.try_get_context` is used in the `app.py` scope to ensure these two keys have values otherwise it exists. I've also made the spot fleet optional with USE_SPOT_INSTANCE, set to true by default

alexiswl:
Just pushed them

reisingerf:
Correct me if I am wrong, but context variables don't have to be specified in `cdk.json`. I think you could just remove them from that file. They can still be added on the command line and you should end up with the same behaviour, no?

alexiswl:
That's true, I like having them in there to inform the user of what context keys are available to them, but maybe that should be highlighted in the readme instead

reisingerf:
Yup. I think users are more likely to read the README than to scan random files in the repo (assuming a standard user does not need to know much about CDK, just how to use it.
Besides, the customisation via the context variables would have to be described in the README anyway and impatient users would face a descriptive error message ;-)

brainstorm:
Great job Alexis! Can't wait to see people using this one more over the Cloudformation one ;)

reisingerf:
Would you prefer this over the CF one for normal users?
All you need for CF is the AWS CLI, which everyone has... 
CDK is a bit more to setup and maintain....

brainstorm:
I would encourage use of CDK instead of CF, yes.

It is more high level, more intuitive, there are examples available and can be written in a familiar language (Python).... just requires python and node installed, which most of our users/devs already have ;)

PR #34, opened 2020-03-31
Dev worker cdk (last minute changes)

alexiswl:
Few minor updates from #33. Added a jupyter notebook with a few shortcuts.


cdk/apps/dev_worker/README.md
```diff
@@ -30,34 +30,101 @@ Since we run through the Fn.Sub to substitute variables into the user_data scrip
...
 ```bash
-cdk diff -c "STACK_NAME=alexis-unique-stack" -c "KEY_NAME=myname-dev' -c "EC2_TYPE=t2.micro"
-cdk deploy -c "STACK_NAME=alexis-unique-stack" "KEY_NAME=myname-dev' -c "EC2_TYPE=t2.micro"
+cdk diff -c "STACK_NAME=alexis-unique-stack" -c "KEY_NAME=myname-dev" -c "EC2_TYPE=t2.micro"
+cdk deploy -c "STACK_NAME=alexis-unique-stack" -c "KEY_NAME=myname-dev" -c "EC2_TYPE=t2.micro"
 ```
 
 Notes:
-If you create a stack with a different *STACK_NAME* parameter, you will need to use this same context parameter when destorying the stack
+You will need to use this same STACK_NAME parameter when destroying the stack
```
reisingerf: Is that really needed? 
I thought CDK creates a `cdk.context.json` with the current context. I would have hoped that a `cdk destroy` would take the context from that file without requiring the parameter again.

alexiswl: I think you're right, but when I wrote this there was a default value for the STACK_NAME in `cdk.json` which was overwriting this hence it needed to be explicitly defined. I will re-test and see if the `try_get_context` method will also pick up from the `cdk.context.json`.

alexiswl: If I run `cdk destroy` without the context parameter STACK_NAME, the `try_get_context` fails to retrieve it. Furthermore, from my directory, the `cdk.context.json` is not updated after running `cdk deploy -c "STACK_NAME=new_instance_name"

reisingerf: Is the 'cdk.context.json` removed with `cdk destroy`?
Trying to figure out if perhaps the context is only generated on the first deploy and kept static...

alexiswl: The `cdk.context.json` file in my directory hasn't been updated since March 3? It contains only the network information such as VPC ID, subnet ID and availability zone

reisingerf: If you `cdk destroy`, then remove the `cdk.context.json` and `cdk deploy` again, does it show the same behaviour or does it generate the correct/updated context file?

alexiswl:
Thanks for summarising these!

> 1. Aim for having no `-c` flags needed when `cdk deploy` and `cdk destroy`.

Would having a uuid for the stack name be acceptable?

> 2. Test if m5.X instances work, not just m4.X. Those have different EBS-OS volume mapping, as in /dev/xvdhX vs /dev/sdaX.

Can develop a workaround if it fails.

> 3. Check if the instance has public IP address(es) after instantiation

I can assume no, but will investigate this further. All ports (including 22) are closed by default.

> 5. Bonus: Verify that the AWS CLI ssm-manager-plugin (and the instance's SSM agent) works with the following (bash/zsh) function:

Is the aim that you can run `ssm ${instance_id} /path/to/a/script`? And see the stdout/stderr in the terminal?

reisingerf:
A few comments from my side...

I like a verbose README, especially if it's like in this case for others to learn from and get introduced. However I agree that a `Quickstart` section on the top would be useful.

There are a few statements in there that should be looked at:
> Our output returns a public IP of our ec2 instance that we can then ssh into via the ssm manager.

According to @brainstorm there is no public IP assignment and if we go over SSM we don't need it, nor port 22.

The `Observe the cdk.json` section could be merged with the `Context parameters` section, which now needs updating as `LAUNCH_TEMPLATE_NAME` is no longer present.

>> 1. Aim for having no -c flags needed when cdk deploy and cdk destroy.

>Would having a uuid for the stack name be acceptable?

I am not a big fan of non-descriptive IDs in stack names. We had a discussion before about this and I thought we came to a conclusion?
Didn't we say to use the context variable if provided and generate a unique one including a default stack name and the $USER env variable (plus possibly a unique string) if not, and only fail with an error if $USER is not defined either)? That should solve the issue.

> 2. Test if m5.X instances work, not just m4.X. Those have different EBS-OS volume mapping, as in /dev/xvdhX vs /dev/sdaX.

I think AWS will generate a link, so both would work if they are not matching the instance default, but definitely worth testing.

> 3. Check if the instance has public IP address(es) after instantiation

Why do we need public IPs?

> 5. Bonus: Verify that the AWS CLI ssm-manager-plugin (and the instance's SSM agent) works with the following (bash/zsh) function

I think that goes back to the discussion on Slack. @brainstorm wanted to use the `--document-name AWS-StartInteractiveCommand` parameter in the AWS CLI command, which requires the latest version of the SSM agent being installed on the instance. He has set up an auto-updating rule, so agents of all instances get automatically updated, but that hasn't been verified yet.

alexiswl:
> Our output returns a public IP of our ec2 instance that we can then ssh into via the ssm manager.
Yep this statement is wrong, it returns the instance ID, which we can then ssh into.

> $USER env variable (plus possibly a unique string)
Yes, I should have been more verbose, this is what I had in mind for the uuid when the parameter is not defined.

> He has set up an auto-updating rule, so agents of all instances get automatically updated, but that hasn't been verified yet.
Could this be added in the user-data?

alexiswl:
> 1. Aim for having no -c flags needed when cdk deploy and cdk destroy.
Done! STACK_NAME created using user and user_id, STACK_NAME added to cdk.json 'manually' by the python script at the end of the deployment which means that `cdk destroy` does not need any context parameters. STACK_NAME can be reset by deleting it from the `cdk.json` file. `cdk context --reset STACK_NAME` does not appear to work. Running with cdk deploy -c "STACK_NAME=foo' will NOT update the stack name in the cdk.json file.

> 2. Test if m5.X instances work, not just m4.X. Those have different EBS-OS volume mapping, as in /dev/xvdhX vs /dev/sdaX.

Yes, these work fine, with the `sd` prefix mount points.

> 3. Check if the instance has public IP address(es) after instantiation
No public IP is present, no ports are open

> 4. Merge @reisingerf commit from master: 86d0e32
Done.

> 5. Bonus: Verify that the AWS CLI ssm-manager-plugin (and the instance's SSM agent) works with the following (bash/zsh) function:

Yes it works, but I prefer the ssh proxy. This command puts you in '/usr/bin' instead of $HOME .

cdk/apps/dev_worker/README.md
```diff
@@ -5,15 +5,25 @@ Deploy an ec2 instance with docker ready to go and logged into our ECR (Elastic
...
+## Quick start:
+1. Create the venv
+`python3 -m venv .env`
+2. Activate the venv
+`source .env/bin/activate`
+3. Deploy the ec2 instance:
+`cdk deploy`
+4. Log into the instance:
+*You will need to read below to correctly configure your ssh config.*
+`ssh i-....`
```
brainstorm, line 17: I would prefer putting SSM instead of SSH in the quickstart.

cdk/apps/dev_worker/app.py
```diff
@@ -15,39 +18,51 @@
...
+
+if update_stack_name:
+    # Set the stack name in the context
+    # Read in
+    with open('cdk.json', 'r') as cdk_fh:
+        cdk_context = json.load(cdk_fh)
+    # Update
+    cdk_context['context']["STACK_NAME"] = stack_name
+    # Write
+    with open("cdk.json", 'w') as cdk_fh:
```
brainstorm, line 66: Nice hack, let's hope there are no side effects with this :)

cdk/apps/dev_worker/cdk.json
```diff
@@ -1,14 +1,12 @@
...
   "context": {
-    "STACK_NAME": null,
     "EC2_TYPE": "m4.4xlarge",
-    "KEY_NAME": null,
     "MACHINE_IMAGE": "ami-0dc96254d5535925f",
     "VAR_VOLUME_SIZE": "16",
     "EXTENDED_VOLUME_SIZE": "100",
     "INSTANCE_NAME": "dev_worker_instance_cdk",
-    "USE_SPOT_INSTANCE": true,
+    "USE_SPOT_INSTANCE": "true",
```
brainstorm, line 9: CDK cross-language data type system forces you to do this I guess?

brainstorm:
@alexiswl Feel free to test and merge my recently pushed changes into this PR: https://github.com/umccr/infrastructure/commit/cac9fad23600ee696de1df90f2d156fbc0d47472

cdk/apps/dev_worker/README.md
alexiswl: Will this allow rsyncing or port binding? I guess we could also write an alias for port binding too with something from here: https://aws.amazon.com/blogs/aws/new-port-forwarding-using-aws-system-manager-sessions-manager/

alexiswl: The rsyncing is just the iap configuration file I like to throw in so I can download and run things from gds on the volume. Not necessary and still possible, ssh isn't ruled out.

brainstorm: Yes, SSH is still in place, but I would like users to start getting more acquainted with SSM since it's easier for us to handle from a security/threat perspective: much better integrated with AWS.

So yes, I would keep the SSH notes down below yet indicate that SSM is preferred though (and leave SSM as the canonical way on the quickstart since most users will follow those 5 steps).

cdk/apps/dev_worker/app.py
alexiswl: The main problem is that cdk.json is part of the git controlled files. Whether it would read the STACK_NAME from cdk.context.json or `~/.cdk.json`, these may be more appropriate alternatives?

cdk/apps/dev_worker/README.md
alexiswl: I don't really see the difference. The ssh is just using ssm underneath. Are we adding in vulnerabilities with the key pairs?

brainstorm: No, look at this: https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-logging-auditing.html.

If there's a breach we can track it down and evaluate the extent of the damage, with SSH you'd have to instrument quite a few more things (logservers, etc...).

alexiswl: So the ProxyCommand in the ssh config is doing something more magical than just a subprocess?

brainstorm: Sorry, I see what you mean and I understand now, both options are fine.

PR #37, opened 2020-04-09
Dev worker cdk

alexiswl:
Last merge didn't add these final commits (including 9eea9e525f10b674402d17996ffd0215b13ed970)

### DIFF
--- a/cdk/apps/dev_worker/README.md
+++ b/cdk/apps/dev_worker/README.md
@@ -137,3 +137,18 @@ cat nohup.out
 ```
 
 And open up a browser tab of the notebook.
+
+## Troubleshooting
+
+### Expired Token Exception
+> An error occurred (ExpiredTokenException) when calling the StartSession operation: The security token included in the request is expired
+ssh_exchange_identification: Connection closed by remote host
+
+This means you need to re-login to aws via the google TFA authentication.
+
+### Target not connected
+> An error occurred (TargetNotConnected) when calling the StartSession operation: i-07aa0098c62dc5001 is not connected.
+ssh_exchange_identification: Connection closed by remote host
+
+The instance takes time to load up, it may not be at the stage that it's ready to accept logins or is yet to populate the authorized keys.
+
--- a/cdk/apps/dev_worker/README.md
+++ b/cdk/apps/dev_worker/README.md
@@ -15,7 +15,7 @@ ssm() {
     instance_id="$1"
     aws ssm start-session --target "${instance_id}" \
                           --document-name "AWS-StartInteractiveCommand" \
-                          --parameters command="sudo su - ec2-user"
+                          --parameters command="sudo su - ssm-user"
 }
 ```
 
@@ -108,13 +108,14 @@ cdk destroy -c "STACK_NAME=alexis-unique-stack"
 
 ### Set SSH config
 This assumes you have loaded your public key complement `aws.pub` into your github keys (which are publicly accessible)
-You may choose between ec2-user and ssm-user, both have identical setups. I would recommend the ec2-user  
+You may choose between ec2-user and ssm-user, both have identical setups. I would recommend the ssm-user.
+*The ec2-user may be removed in the near future if all the ssm functionality works correctly*
 You will need to use the ssh command over ssm if you wish to rsync data from your local machine to the instance.  
 Note that data transfer is very slow as we're not going over port 22 directly.
 ```
 host i-* mi-*
     ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
-    User ec2-user
+    User ssm-user
     IdentityFile /path/to/private/key/.ssh/aws
 ```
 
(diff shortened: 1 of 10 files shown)


cdk/apps/dev_worker/README.md
```diff
@@ -60,7 +66,7 @@ Context parameters are defined with the `-c "STACK_NAME=my-stack` argument.
 Repeat the `-c` parameter for defining multiple parameters.    
 i.e 
 ```
-cdk deploy -c "STACK_NAME=my-stack"
+cdk deploy -c "STACK_NAME=my-stack" -c "USE_SPOT_INSTANCE=false"
```
brainstorm, line 69: @alexiswl Why? SPOT instances are **very** much preferred, by default :)

cdk/apps/dev_worker/README.md
```diff
@@ -101,18 +108,45 @@ cdk destroy -c "STACK_NAME=alexis-unique-stack"
...
 ```
 host i-* mi-*
     ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
     User ec2-user
     IdentityFile /path/to/private/key/.ssh/aws
 ```
 
+### Setting port forwarding through SSM
+
+Set up the following shell function in your `~.bashrc`.
```
brainstorm, line 123: @alexiswl Fantastic! <3

cdk/apps/dev_worker/user_data/user_data.sh
```diff
@@ -128,4 +128,12 @@ rm Anaconda3-${!ANACONDA_VERSION}-Linux-x86_64.sh
 
 # Fix bashrc for ec2-user and ssm-user for access ready for conda
 su - "ec2-user" -c "/opt/conda/bin/conda init"
-su - "ssm-user" -c "/opt/conda/bin/conda init"
\ No newline at end of file
+su - "ssm-user" -c "/opt/conda/bin/conda init"
+
+# Pull starter notebook into home directory
+# FIXME update notebook path once in master branch
+starter_notebook="https://raw.githubusercontent.com/alexiswl/umccr-infrastructure/dev_worker_cdk/cdk/apps/dev_worker/dev_worker_notebook.ipynb"
```
brainstorm, line 135: Brilliant! :)

cdk/apps/dev_worker/README.md
reisingerf: Happy for it to be configurable, as long as the default is SPOT.
However, I would remove the spot price parameter. It's kind of annoying to first find out what the current spot price is and then adjust it each time you want to change the instance type.

alexiswl:
So the m5 doesn't work just yet, my guess is there's problems with the eb2 volumes mount. Let me re-look into this.

brainstorm:
Weird, it did work when I tested it myself :-S What's failing on your end?

alexiswl:
> Weird, it did work when I tested it myself :-S What's failing on your end?

Not sure, and working fine through a new ami type - the current one is about ten months old. It could have actually been from not running cdk destroy beforehand and having left over stack problems. Few other teething issues I've noticed with the userdata that I'll need to fix up before merging.

cdk/apps/dev_worker/README.md
alexiswl: Can't pause a spot instance and every now and again they go away which can be annoying. I think for the dev worker use case there's definitely a need for non-spot instances.

alexiswl: A default spot price parameter needs to be set though right?

reisingerf: You don't need a spot price parameter (IIRC), only setting the `MarketType` to `spot` should use the default spot price (which is usually the on-demand price).

alexiswl: So one should never need to set the spot price? Or should this remain as an option?

alexiswl:
Also I don't like keeping the "STACK_NAME" inside the cdk.json file, it syncs with git by default and I find myself performing continuous soft resets. Thoughts on having a file called .stack_name which is read in / written by cdk? And we can add this to .gitignore?

cdk/apps/dev_worker/README.md
reisingerf: I would vote for the former. The less effort the better and setting a spot price explicitly is only useful in very rare cases.

reisingerf:
TBH I would keep it as simple as possible and keep a user specific default stack name, as previously suggested. 
There should not be any need for cdk.json hacks and it should work for the majority of use cases.

alexiswl:
> TBH I would keep it as simple as possible and keep a user specific default stack name, as previously suggested.
> There should not be any need for cdk.json hacks and it should work for the majority of use cases.

I've amended this as so.

1. A file called .stack_name (which is part of git ignore) is created in the case that a uuid stack is used.

The hierachy is as follows.
Context (-c) -> .stack_name -> new uuid.

If a stack name is provided as part of the context parameter, the .stack_name is not updated. To refresh a new uuid stack_name, the .stack_name file can be deleted and a new one will created on the next cdk command.

brainstorm:
@alexiswl Will that `.stack_name` will be kept on the current repository? $HOME? Elsewhere?

As long as it doesn't produce confusion among users I'm fine as it is ;)

alexiswl:
Just in the same directory as cdk.json and cdk.context.json. I added the file to the .gitignore too.

brainstorm:
Alrighty, shall we merge as-is and let @pdiakumis give it a go (he volunteered as pro end-user)? ;)

PR #49, opened 2020-06-08
Allowed DNS hostname in networking stack main vpc

victorskl:
* Usually DNS hostname is desirable
* Updated README to get this main vpc by name and tags filter
  Reduce attack surface by making lesser known VPC ID
  exposure in downstream terraform and CDK code

### DIFF
--- a/terraform/stacks/networking/README.md
+++ b/terraform/stacks/networking/README.md
@@ -19,17 +19,18 @@ At the moment, this VPC defines three "tiers" of subnets:
 
 #### Terraform
 
-- In your app terraform, you can use [data source: aws_vpc](https://www.terraform.io/docs/providers/aws/d/vpc.html) to query this VPC by its ID.
+- In your app terraform, you can use [data source: aws_vpc](https://www.terraform.io/docs/providers/aws/d/vpc.html) to query this VPC. Recommended to use tag filters on `Name`, `Stack` and `Environment` keys over VPC ID as example below.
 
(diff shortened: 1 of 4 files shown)


terraform/stacks/networking/README.md
```diff
@@ -72,28 +82,30 @@ from stacks.common import CommonStack
 
 app = core.App()
 
-common_props = {
-    'vpc_id': 'vpc-01234567890'  # output of main vpc ID
-}
-
-batch_app_props = {}
+env_dev = core.Environment(account="123456789", region="ap-southeast-2")
```
brainstorm, line 85: One "principle" we discussed with @reisingerf recently is that we'll start taking things like accountIDs away from (CDK) scripts, fetching them implicitly from the currently authenticated account (on `~/.aws`, parameter store, `cdk.context.json` or similar). I'm not 100% sure if that's easy/doable with this particular example, but just leaving it here as an observation.

victorskl: - Yap, agree. Updated to use those from `~/.aws` i.e. with `cdk synth --profile=dev` [flag from AWS CLI profile](https://docs.aws.amazon.com/cdk/latest/guide/environments.html).
- Please see updated [README.md](https://github.com/umccr/infrastructure/tree/networking-dns-hostname/terraform/stacks/networking) on this.
- Also provided [`debug.py`](https://github.com/umccr/infrastructure/blob/networking-dns-hostname/terraform/stacks/networking/debug.py) sample for CDK try out for querying this main vpc and its subnets tier filtering.

PR #53, opened 2020-07-06
Samplesheet by index len

alexiswl:
This adjusts the samplesheet-split script with the following changes for bclconvert compatibility.

1. Samplesheets split by sample type AND also by index lengths.
2. Trailing Ns in indexes removed.
3. Additional parameters `--all`, `--sample-type`, `--index-length`, `--index2-length` added.

Use `--all` to keep previous logic, which creates a samplesheet for each combination of sample type and index length with the naming convention `SampleSheet_<sample_type>.<index_length>.<index2_length>.csv`.

Use `--sample-type` and `--index-length` and `--index2-length` to specifiy a samplesheet at the location `--output-path`.
The defaults of `--sample-type`, `--index-length` and `--index2-length` are 'WGS', '8' and '8' respectively.

### DIFF
--- a/scripts/showcase/samplesheet-prep-test.py
+++ b/scripts/showcase/samplesheet-prep-test.py
@@ -152,6 +152,8 @@ def read_sample_sheet(sample_sheet):
             Sample_Type         What type of sample is this - eg DNA (`str) - optional
             Pair_ID             Which pair of samples does this belong to - eg est_Sample_UP01 -
                                 this may be the same as Sample_ID - required
+            index_len           The length of the first index - used in groupings
+            index2_len          The length of the second index - using in groupings
             ==========  ==============================================================
     """
     # Find the pivot (The row that contains [Data] that splits the samplesheet from the header
@@ -173,6 +175,14 @@ def read_sample_sheet(sample_sheet):
 
     # Re-read the sample sheet
     sample_sheet_df = pd.read_csv(sample_sheet, skiprows=start_row+1, header=0)
+
+    # Strip Ns at the end of indexes
+    sample_sheet_df['index'] = sample_sheet_df['index'].apply(lambda x: x.rstrip("N"))
+    sample_sheet_df['index2'] = sample_sheet_df['index2'].apply(lambda x: x.rstrip("N"))
+
+    # Add lengths
+    sample_sheet_df['index_len'] = sample_sheet_df['index'].apply(lambda x: len(x))
+    sample_sheet_df['index2_len'] = sample_sheet_df['index2'].apply(lambda x: len(x))
 
     return sample_sheet_header_rows, sample_sheet_df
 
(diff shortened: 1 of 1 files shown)


scripts/showcase/samplesheet-prep-test.py
```diff
@@ -68,9 +68,26 @@ def get_args():
...
+    parser.add_argument("--sample-sheet", "--samplesheet", "-s", type=str, required=True, dest="sample_sheet")
+    parser.add_argument("--tracking-sheet", "--trackingsheet", "-t", type=str, required=True, dest="tracking_sheet")
+    parser.add_argument("--output-path", "--outputPath", "-o", type=str, required=True, dest="output_path")
+
+    sample_params_args = parser.add_argument_group("Parameters for filtering out samples."
+                                                   "If --sample-type, --index-legnth or --index2length"
+                                                   "are used. A file named SampleSheet.csv will be output."
+                                                   "If the --all parameter is used, then each file named will be "
+                                                   "distinguished with the following notation"
+                                                   "SampleSheet_<sample_type>.<index-length>.<index2-length>.csv")
```
skanwal, line 80: If I understand this correctly, we had agreed to continue generating `SampleSheet.csv` (which by default would be of type WGS and index length (8,8) - if these indexes exist or (8,6) otherwise? For this case, I'll need to update CWL tool definition by adding ` --sample-type, --index-legnth & --index2length` parameters. If I set a default value for these and not pass any input parameter at workflow launch stage - would it still default to generating `SampleSheet.csv`?

scripts/showcase/samplesheet-prep-test.py
```diff
@@ -403,17 +441,48 @@ def write_sample_sheets(sample_sheet_header_rows, sample_sheet_df, sample_sheet_
...
         # Modify the sample-sheet
         modified_sample_header_rows, modified_sample_sheet_df = modify_sample_sheet(
             sample_sheet_header_rows=sample_sheet_header_rows,
             sample_sheet_df=sample_sheet_type_df,
             sample_type=sample_type)
 
-        # Set the output file name
-        output_file = sample_sheet_dir / "SampleSheet_{}.csv".format(sample_type)
+        if sample_type is not None and index_len is not None and index2_len is not None:
+            output_file = sample_sheet_dir / "SampleSheet.csv"
```
skanwal, line 480: This suggests, the user needs to pass these _three_ arguments to produce `SampleSheet.csv` file - which will be based on specified inputs?

scripts/showcase/samplesheet-prep-test.py
alexiswl: Defaults of `sample-type`, `index-length` and `index2length` are `WGS`, `8` and `8`, so you don't have to set any defaults as they're set by the script although you may wish to so that CWL users can see them easier.

If you want to extract the samplesheet for WGS samples with 8 and 6 you would need to set `--index2-length` to '2'
If you don't, this will terminate with exit code 1 since no samples would be found (if no samples have the index 8, 8)

One main change is that it will generate the file `Samplesheet.csv` rather than `SampleSheet_<sample_type>.csv`

scripts/showcase/samplesheet-prep-test.py
alexiswl: These are set by default, if the `-all` parameter is set, these values are set to none and all possible samplesheets grouped by type, index_len, and index2_len are generated.

alexiswl: all_arg = getattr(args, 'all')
    if all_arg:
        logger.info("Setting sample-type, index and index2 all to none since the --all parameter was specified")
        setattr(args, "sample_type", None)
        setattr(args, "index_length", None)
        setattr(args, "index2_length", None)

    return args

scripts/showcase/samplesheet-prep-test.py
skanwal: - `--index-legnth` - typo needs fixing
- both index argument names need to be harmonised

scripts/showcase/samplesheet-prep-test.py
skanwal: That sounds good!

PR #56, opened 2020-07-08
Added override cycles logic

alexiswl:
--no-override-cycles parameter to stop this logic from processing.

Otherwise, by default adds Override cycles to [Settings] block in SampleSheet.csv.

If the index has trailing Ns, the index is stripped to that length and the rest is populated with Ns,
i.e
an index of length 8 when 10 cycles have been run will present as:
i8n2

Reads are taken from the [Reads] block for R1 and R2.

Also fixed bug where index_len and index2_len columns were being added to the output samplesheet.csv file.

### DIFF
--- a/scripts/showcase/samplesheet-prep-test.py
+++ b/scripts/showcase/samplesheet-prep-test.py
@@ -152,6 +152,8 @@ def read_sample_sheet(sample_sheet):
             Sample_Type         What type of sample is this - eg DNA (`str) - optional
             Pair_ID             Which pair of samples does this belong to - eg est_Sample_UP01 -
                                 this may be the same as Sample_ID - required
+            index_len           The length of the first index - used in groupings
+            index2_len          The length of the second index - using in groupings
             ==========  ==============================================================
     """
     # Find the pivot (The row that contains [Data] that splits the samplesheet from the header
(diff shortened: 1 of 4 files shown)


scripts/showcase/samplesheet-prep-test.py
```diff
@@ -227,12 +239,38 @@ def convert_sample_sheet_header_to_v2(sample_sheet_header_rows):
...
+            if not index2_len == index2_len_orig:
+                # We need to mask the length of the difference
+                i5_mask = "{}i{}n".format(index2_len, index2_len_orig-index2_len)
+            else:
+                i5_mask = "{}i".format(index2_len)
+
+            sample_sheet_header_rows_new.append("OverrideCycles,{};{};{};{}\n".format(
+                r1_mask, i7_mask, i5_mask, r2_mask
+            ))
         sample_sheet_header_rows_new.append(new_row)
```
skanwal, line 272: An example of a correct format for overridecycles setting for bclConvert is:
`OverrideCycles,Y151;I8N2;I8N2;Y151`

You are adding:
OverrideCycles,151y;8i2n;8i2n;151y

alexiswl: :facepalm:

alexiswl: This should be fixed now with #57

alexiswl:
@skanwal can we merge this.

skanwal:
Sorry - forgot to press the button after testing. Thanks.

PR #59, opened 2020-07-10
Networking stack enable VPC Gateway Endpoints

victorskl:
* This allows data communication for S3 and DynamoDB
  without going through NAT Gateway


victorskl:
Abandoning this PR. Will create another PR again.

PR #64, opened 2020-08-10
Init commit to parallel_cluster

alexiswl:
Re-configured base.yml and conda.yml
Added args and checks to start_cluster.sh
Created umccr_dev configuration with new ami (which has base.yml and conda.yml) and uses lustre_fs
Wrote fixme under environment, can't login and run in the same env.

Still having some issues creating the system however:

<details>

<summary>Last 100 lines of `/var/log/cfn-init.log`</summary>

```bash
  ================================================================================
  Recipe Compile Error in /etc/chef/local-mode-cache/cache/cookbooks/aws-parallelcluster/recipes/slurm_config.rb
  ================================================================================
  
  Mixlib::ShellOut::ShellCommandFailed
  ------------------------------------
  Expected process to exit with [0], but received '127'
  ---- Begin output of /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws fsx --region ap-southeast-2 describe-file-systems --file-system-ids fs-07a9895a51552e844 --query 'FileSystems[0].LustreConfiguration.MountName' --output text ----
  STDOUT: 
  STDERR: sh: /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws: No such file or directory
  ---- End output of /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws fsx --region ap-southeast-2 describe-file-systems --file-system-ids fs-07a9895a51552e844 --query 'FileSystems[0].LustreConfiguration.MountName' --output text ----
  Ran /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws fsx --region ap-southeast-2 describe-file-systems --file-system-ids fs-07a9895a51552e844 --query 'FileSystems[0].LustreConfiguration.MountName' --output text returned 127
  
  Cookbook Trace:
  ---------------
    /etc/chef/local-mode-cache/cache/cookbooks/aws-parallelcluster/recipes/fsx_mount.rb:35:in `from_file'
    /etc/chef/local-mode-cache/cache/cookbooks/aws-parallelcluster/recipes/base_config.rb:78:in `from_file'
    /etc/chef/local-mode-cache/cache/cookbooks/aws-parallelcluster/recipes/slurm_config.rb:18:in `from_file'
  
  Relevant File Content:
  ----------------------
  /etc/chef/local-mode-cache/cache/cookbooks/aws-parallelcluster/recipes/fsx_mount.rb:
  
   28:      owner 'root'
   29:      group 'root'
   30:      mode '1777'
   31:      action :create
   32:    end
   33:  
   34:    require 'chef/mixin/shell_out'
   35>>   mountname = shell_out!("#{node['cfncluster']['cookbook_virtualenv_path']}/bin/aws fsx --region #{node['cfncluster']['cfn_region']} describe-file-systems --file-system-ids #{node['cfncluster']['cfn_fsx_fs_id']} --query 'FileSystems[0].LustreConfiguration.MountName' --output text", :user=>'root').stdout.strip
   36:  
   37:    mount_options = %w[defaults _netdev flock user_xattr noatime]
   38:  
   39:    mount_options.push(%w[noauto x-systemd.automount]) if node['init_package'] == 'systemd'
   40:  
   41:    # Mount FSx over NFS
   42:    mount fsx_shared_dir do
   43:      device "#{node['cfncluster']['cfn_fsx_fs_id']}.fsx.#{node['cfncluster']['cfn_region']}.#{node['cfncluster']['aws_domain']}@tcp:/#{mountname}"
   44:      fstype 'lustre'
  
  System Info:
  ------------
  chef_version=15.11.8
  platform=amazon
  platform_version=2
  ruby=ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]
  program_name=/opt/cinc/bin/cinc-client
  executable=/opt/cinc/bin/cinc-client
  
  
  Running handlers:
[2020-08-10T12:52:17+00:00] ERROR: Running exception handlers
  Running handlers complete
[2020-08-10T12:52:17+00:00] ERROR: Exception handlers complete
  Cinc Client failed. 3 resources updated in 04 seconds
[2020-08-10T12:52:17+00:00] FATAL: Stacktrace dumped to /etc/chef/local-mode-cache/cache/cinc-stacktrace.out
[2020-08-10T12:52:17+00:00] FATAL: Please provide the contents of the stacktrace.out file if you file a bug report
[2020-08-10T12:52:17+00:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '127'
---- Begin output of /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws fsx --region ap-southeast-2 describe-file-systems --file-system-ids fs-07a9895a51552e844 --query 'FileSystems[0].LustreConfiguration.MountName' --output text ----
STDOUT: 
STDERR: sh: /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws: No such file or directory
---- End output of /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws fsx --region ap-southeast-2 describe-file-systems --file-system-ids fs-07a9895a51552e844 --query 'FileSystems[0].LustreConfiguration.MountName' --output text ----
Ran /opt/parallelcluster/pyenv/versions/3.6.9/envs/cookbook_virtualenv/bin/aws fsx --region ap-southeast-2 describe-file-systems --file-system-ids fs-07a9895a51552e844 --query 'FileSystems[0].LustreConfiguration.MountName' --output text returned 127

2020-08-10 12:52:17,779 [ERROR] Error encountered during build of chefConfig: Command chef failed
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/construction.py", line 542, in run_config
    CloudFormationCarpenter(config, self._auth_config).build(worklog)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/construction.py", line 260, in build
    changes['commands'] = CommandTool().apply(self._config.commands)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/command_tool.py", line 117, in apply
    raise ToolError(u"Command %s failed" % name)
ToolError: Command chef failed
2020-08-10 12:52:17,781 [ERROR] -----------------------BUILD FAILED!------------------------
2020-08-10 12:52:17,781 [ERROR] Unhandled exception during build: Command chef failed
Traceback (most recent call last):
  File "/opt/aws/bin/cfn-init", line 171, in <module>
    worklog.build(metadata, configSets)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/construction.py", line 129, in build
    Contractor(metadata).build(configSets, self)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/construction.py", line 530, in build
    self.run_config(config, worklog)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/construction.py", line 542, in run_config
    CloudFormationCarpenter(config, self._auth_config).build(worklog)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/construction.py", line 260, in build
    changes['commands'] = CommandTool().apply(self._config.commands)
  File "/usr/lib/python2.7/site-packages/cfnbootstrap/command_tool.py", line 117, in apply
    raise ToolError(u"Command %s failed" % name)
ToolError: Command chef failed
2020-08-10 12:52:18,032 [DEBUG] CloudFormation client initialized with endpoint https://cloudformation.ap-southeast-2.amazonaws.com
2020-08-10 12:52:18,033 [DEBUG] Signaling resource MasterServerWaitCondition in stack parallelcluster-AlexisFirstCluster with unique ID i-0b954110d7716974f and status FAILURE
```
</details>


parallel_cluster/conf/config
```diff
@@ -37,8 +65,9 @@ shared_dir       = efs
...
-#shared_dir                = /scratch
-#storage_capacity          = 3600
-#imported_file_chunk_size  = 1024
-#import_path               = s3://tothill-scratch
+[fsx lustrefs]
+shared_dir                = /fsx
+storage_capacity          = 3600
+imported_file_chunk_size  = 1024
+import_path               = s3://umccr-temp-dev/Alexis_parallel_cluster_test/input_folder
+export_path               = s3://umccr-temp-dev/Alexis_parallel_cluster_test/output_folder
```
brainstorm, line 73: After our tests w/ Florian we decided to leave `export_path` out since it's just confusing (must use `hsm_upload` tool instead of just assuming that changes on the filesystem/bucket are transparent).

parallel_cluster/conf/pcluster_client.env.yml
```diff
@@ -4,6 +4,9 @@ dependencies:
   - pip
   - boto3
   - conda-forge::jq
+  # FIXME this install awscli v1.18
```
brainstorm, line 7: True, this should be removed and superseeded by EC2 Image Builder component `awscliv2` (or amazon-linux-extras equivalent if there's one), making it system-wide instead of conda/user dependant.

parallel_cluster/README.md
```diff
@@ -31,9 +31,14 @@ $ ssm i-XXXXXXXXXXXXX
 ```
 
 ### Running the cluster
+> CLUSTER_NAME can only be alpha-numeric and must start with a letter
 
 ```shell
-$ ./bin/start-cluster.sh <CLUSTER_NAME>
```
brainstorm, line 36: I just wanted to have devs/users instantiate a cluster fast without caring about those long flags (by using reasonable defaults, only requiring a cluster name) :_S

parallel_cluster/bin/start_cluster.sh
```diff
@@ -1,21 +1,87 @@
...
+      ;;
+    --extra-parameters)
+      extra_parameters_arg="$2"
+      shift 1
+      ;;
+    --help)
+      display_help
+      exit 0
+      ;;
+    *)
```
brainstorm, line 38: We can have the cake and it it too: If it doesn't begin with `--`, assume it's a clustername and Bob's your uncle?

parallel_cluster/bin/start_cluster.sh
```diff
@@ -1,21 +1,87 @@
...
+	  if [[ "$?" == "0" ]]; then
+      echo_stderr "Stack creation succeeded - now retrieving name of IP"
+      exit 0
+    else
+      echo_stderr "Could not create the parallel cluster stack, exiting"
+      exit 1
+    fi
+
+    # Output the master IP to log
+    echo_stderr "$(aws ec2 describe-instances \
```
brainstorm, line 77: Don't recall if this actually worked in practice?

parallel_cluster/conf/config
```diff
@@ -23,12 +23,40 @@ scheduler             = slurm
...
 use_public_ips = true
 
+[vpc umccr_dev_network]
+# Default vpc for dev account
+vpc_id = vpc-00eafc63c0dfca266
+# Compute subnet ids take the same subnet as the master
+# Our default public subnet
+master_subnet_id = subnet-0fab038b0341872f1
+# Elastic IP address is associated to the master instance.
+use_public_ips = true
```
brainstorm, line 58: We might want to `false` this since users should login via SSM.

parallel_cluster/README.md
```diff
@@ -31,9 +31,14 @@ $ ssm i-XXXXXXXXXXXXX
...
+$ CLUSTER_TEMPLATE="tothill"  # or umccr_dev 
+$ ./bin/start_cluster.sh \
+  --cluster-name <CLUSTER_NAME> 
+  --cluster-template "${CLUSTER_TEMPLATE}"
+  --extra-parameters "{\"KeyName\": \"alexis-wfh-dev\"}"
```
brainstorm, line 41: On the keyname discussion: please use and encourage SSM instead, we'll be better off if we don't have to track public SSH keys in-AMI or instance (plus we can log all SSM user activity on an S3 bucket and audit properly).

parallel_cluster/conf/config
```diff
@@ -23,12 +23,40 @@ scheduler             = slurm
...
+# 16 CPUs and 64 Gb
+compute_instance_type = m5.4xlarge
+ec2_iam_role          = parallelcluster-ec2-instance-role
+scheduler             = slurm
+initial_queue_size    = 1
+# Basic ami with installations of
+# R, python3.8, conda, pip, ruby, golang, rust
+# conda has been initialised to the ec2-user
+custom_ami            = ami-0b98c49201c9a1e75
+
```
reisingerf, line 44: Can we add `tags`?
(to avoid Config rule violations)

parallel_cluster/bin/start_cluster.sh
alexiswl: Can confirm `pcluster create` does return a non-zero exit code.

parallel_cluster/README.md
alexiswl: Yes but how do they specify which config? And KeyName, which I can discuss further in the SSM comment below.

parallel_cluster/README.md
alexiswl: If I don't place in KeyName in either the config or as extra-parameters, I get an error stating that this attribute is missing. How do we create the instance if we don't have a KeyName to begin with?

parallel_cluster/conf/config
alexiswl: What's the process then in order to map files back to the s3 buckets on closure of the cluster? Is there a way to place this under a separate mount point?

parallel_cluster/README.md
brainstorm: IIRC Florian had separate config files for each cluster? (`tothill`, `dev`, etc...)

parallel_cluster/conf/pcluster_client.env.yml
alexiswl: I think that's possible, with conda env-vars.sh they can prepend aws2 to their path

parallel_cluster/README.md
reisingerf: Yes, I think you need an SSH key, but that does not have to be yours. It can be any that's available in the account and it should (generally) be ignored. I'd probably set a default one in the `config` and forget about it.

parallel_cluster/README.md
reisingerf: I started off having different cluster configurations in the same `config` file, but that may not scale to the number of use cases we'll need to support.
In any case, I'd focus on one use case for now and split it up if/when we need to.

parallel_cluster/README.md
alexiswl: Right and then install ssm into the AMI s.t we can run ssm inside?

brainstorm: I suspect that Amazon Linux 2 (for which I'm basing those EC2 Image Builder components off of) already has SSM reqs installed?

reisingerf: It should yes. You just need to make sure to include the permissions in the instance profile...
I think `AmazonSSMManagedInstanceCore` should do it.

parallel_cluster/conf/config
brainstorm: @alexiswl On the AWS FSx web console, we found that the following checkbox combinations give the closest POSIX filesystem UX:

Screen Shot 2020-08-11 at 2 18 15 pm

... that is, no weird export-specific commands, should just behave as a "close-to-normal" filesystem. There should not be separate mountpoints, just one.

brainstorm: Related AWS (recent) announcement: https://aws.amazon.com/about-aws/whats-new/2020/07/amazon-fsx-lustre-supports-automatic-updates-amazon-s3/

alexiswl:
Looks like it might not be as simple as creating an AMI and running it with ParallelCluster:

Will dig in for these this afternoon:
https://github.com/aws/aws-parallelcluster/issues/1334
https://docs.aws.amazon.com/parallelcluster/latest/ug/tutorials_02_ami_customization.html

parallel_cluster/README.md
alexiswl: I added in AmazonSMMManagedInstanceCore as an additionalPolicy in the config and this enabled me to SSM into the instance.
I've also placed some commands in the base.yml image builder to ensure the latest version of SSM is installed on the image.

parallel_cluster/conf/config
alexiswl: Reading the docs, does this mean we should be using the configuration that expects an existing filesystem instead?

See: https://docs.aws.amazon.com/parallelcluster/latest/ug/fsx-section.html#fsx-fs-id

parallel_cluster/conf/config
alexiswl: This has been added in the start_cluster.sh script with the following commands:
```
      --cluster-template "${cluster_template_arg}" \
      --tags "{\"Creator\": \"${USER}\"}"
```

parallel_cluster/conf/config
alexiswl: Will test early this week to see if there are any unintended consequences.

parallel_cluster/bin/start_cluster.sh
alexiswl: Have seen other cases where this is not the case however. Will leave this conversation open for now.

parallel_cluster/conf/pcluster_client.env.yml
alexiswl: I've got this working by installing just pip and aws-parallel-cluster through pip. Do we need boto3 and jq and awscli?

parallel_cluster/README.md
alexiswl: Keyname now resolved through ssm (do we have a keyname that's a dud - as in no one has the private pair?)
Long flags no longer required.
At the moment only two different configs exist in the config file, I think that's pretty reasonable.

parallel_cluster/conf/config
brainstorm: Yes, in my experience they should be created *a priori*, this is what I had to run on the instance to have them visible:

```
mount -t lustre -o noatime,flock fs-0a1c20d2dfb65e572.fsx.ap-southeast-2.amazonaws.com@tcp:/j6gqfbmv /scratch
mount -t lustre -o noatime,flock fs-0f042be802e974004.fsx.ap-southeast-2.amazonaws.com@tcp:/mwgafbmv /mnt/data
mount -t lustre -o noatime,flock fs-04acac0ac63c53a05.fsx.ap-southeast-2.amazonaws.com@tcp:/jlyafbmv /mnt/refdata
```

Both the part before the `@` as well as the ID coming after `:/` are generated when a FSx filesystem is created (via the webconsole or AWS CLI... in our case, preferably via CDK).

parallel_cluster/conf/pcluster_client.env.yml
brainstorm: Yes, I bet users will be needing all of those, so it'd be nice to have them pre-installed on the AMI/base system, IMHO.

parallel_cluster/conf/config
alexiswl: I see, thanks! Yes, CDK might be the best bet for now - otherwise it's hard-coding them in the post_install args.
You didn't need to include these in the config at all?

brainstorm: No, I didn't need them on the pcluster config on my tests: 

I defined the FSx filesystem via the web console, spawned the pcluster with `--norollback` and then ran the aforementioned mount commands (once the FSx filesystems were ready/created, they can take <5-6min to be ready).

alexiswl: This should be fine, both the container and the /tmpdir for the cromwell job sit on the standard ebs volume. (Cromwell jobs generally put the tmp dir inside the folder for java based jobs but I put it a sed command in the sbatch script to modify the script to set to a root mount point).

alexiswl: I think there would be a benefit to have the refdata entirely on SSD though, it's only a few GB so that shouldn't blow out costs right? Or is it a set cost to have the file system which is at minimum 1.2TBs?

parallel_cluster/conf/pcluster_client.env.yml
alexiswl: So we wait for AWS to create v2 of their pip installation? Should we expect that to be coming soon?

parallel_cluster/conf/config
brainstorm: > I think there would be a benefit to have the refdata entirely on SSD though, it's only a few GB so that shouldn't blow out costs right? Or is it a set cost to have the file system which is at minimum 1.2TBs?

The latter, I had it running for just 2-3 days and it shoot up to >$50.

parallel_cluster/conf/pcluster_client.env.yml
brainstorm: Is it tricky/messy to install v2 from our EC2-Image-Builder components?

parallel_cluster/conf/config
alexiswl: Yep that looks okay then, from https://aws.amazon.com/fsx/lustre/pricing/ *For HDD-based file systems, the optional SSD cache improves performance by automatically placing your most frequently read data on SSD (the cache size is 20% of your file system size).* so all of the ref data would fit inside that 20%.*

However to get that 80% drop they've bragged about here, it does seem to be on the extreme lower end of the HDD use only.

The top HDD baseline (40MB/s/TiB) with 20% SSD cache compared to the lowest SSD-baseline (50MB/s) gives is only a 33% saving.
The lowest HDD baseline (12 MB/s/TiB) with 20% SSD cache compared to the lowest SSD-baseline (50MB/s) gives a 72 % saving.

Reading through this, it seems you do only pay for the filesystem when it's running, obvs storage in S3 that it's mounted to too. If the master instance is stopped, is the filesystem also terminated? I think that could be quite useful for someone that hasn't finished their work but would like to respawn. AFAIK the master instance isn't a spot instance and could be shutdown over night?

brainstorm: > If the master instance is stopped, is the filesystem also terminated?

AFAIK, no, when you define the FSx via the web console, it goes `running` and that's totally independent from any EC2 instance, you can have 0 instances running and the FSx filesystem is "running", even if nothing uses it.

It is a pity that AWS does not charge for *USE* (as in I/O), like they do with S3. But it is understandable, the nature of FSx is essentially the same as EBS: a dedicated portion of a spinning drive (HDD) or memory ICs (SDD).

> AFAIK the master instance isn't a spot instance and could be shutdown over night?

No, [the pcluster `config` file is indeed SPOT](https://github.com/umccr/infrastructure/blob/master/parallel_cluster/conf/config#L16)... both master and compute are IIRC (when I looked at EC2 spot, there were 2 fulfilled spot requests every time I spawned the cluster).

alexiswl: >  FSx filesystem is "running", even if nothing uses it.
Right, but it can be shut down and started upon request?

brainstorm: > Right, but it can be shut down and started upon request?

Yes :)

alexiswl: > No, the pcluster config file is indeed SPOT... both master and compute are IIRC

Looks like the master node can be shut down - suggests its not a spot instance.

image

alexiswl: Also lifecycle attribute is 'normal'

brainstorm:
@alexiswl Are you ok with reviewing and merging this (beefy) PR soon? I would like to prepare and show @AndrewPattison what you have been fixing/cooking here ;)

alexiswl:
> @alexiswl Are you ok with reviewing and merging this (beefy) PR soon? I would like to prepare and show @AndrewPattison what you have been fixing/cooking here ;)

Can you give me the morning to touch up the documentation a bit / test the bcbio conda install works rather than put in something that may well fail to build? - I should have done this before the last commit.
I would like to document how to build the RDS database and how to submit to cromwell too.

Happy to leave the rest of that for a future pull request, which is mainly going to be fsx related.

brainstorm:
Excellent, let's do it like so then 👍🏻

alexiswl:
@brainstorm happy to merge with the changes in eed11c5

parallel_cluster/conf/pcluster_client.env.yml
alexiswl: I thought this the virtualenv installation for the local user to then run pcluster - from https://github.com/aws/aws-cli/issues/4947 it seems clear that they're not going to update it. Comments from users are pretty brutal.

brainstorm (approved):
Fantastic!

alexiswl:
Registered --norollback issue at https://github.com/aws/aws-parallelcluster/issues/1955

PR #68, opened 2020-08-30
Complementary python script inline in the cwl workflow

alexiswl:
Python script to match: https://github.com/umccr-illumina/cwl-iap/pull/130

### DIFF
--- a/scripts/showcase/samplesheet-by-override-cycles.py
+++ b/scripts/showcase/samplesheet-by-override-cycles.py
@@ -1,0 +1,336 @@
+#!/usr/bin/env python3
+
+"""
+Take in a samplesheet,
+Look through headers, rename as necessary
+Look through samples, update override cycles logic as necessary
+Split samplesheet out into separate override cycles files
+Write to separate files
+"""
+
+# Imports
+import re
+import os
+import pandas as pd
+import logging
+import argparse
+from pathlib import Path
+import sys
+
+# Set logging level
+logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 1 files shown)


scripts/showcase/samplesheet-by-override-cycles.py
```diff
@@ -0,0 +1,336 @@
...
+
+    # Arguments
+    parser.add_argument("--samplesheet-csv", required=True,
+                        help="Path to samplesheet csv")
+    parser.add_argument("--out-dir", required=False,
+                        help="Output directory for samplesheets, set to cwd if not specified")
+    parser.add_argument("--samples", required=False,
+                        help="Comma separated values to match override-cycles arg")
+    parser.add_argument("--override-cycles", required=False,
+                        help="Comma separated values to match samples arg")
```
skanwal, line 55: To confirm, if these arguments are "None", the same input samplesheet is returned?

skanwal (approved):
Thanks. One quick question for my context otherwise works great.

scripts/showcase/samplesheet-by-override-cycles.py
alexiswl: Yes, unless `--v2` is also specified, in which case the samplesheet headers are converted for `Settings` and `Data`

alexiswl: And that same samplesheet is placed in the directory specified by `--out-dir` as `SampleSheet.csv`

PR #69, opened 2020-09-01
Rerouted from temp-dev to research dev, built-in AccessFrom to cluster to prevent ssh access from everywhere

alexiswl:
Start_cluster:
Uses AccessFrom to users local IP to restrict access to the cluster from ssh to just them. More for security checks, we don't use ssh access anyway.

Readme:
Added in SSO login issues whilst in pcluster env

All other file changes due to redirection of config files to a new s3 bucket.

### DIFF
--- a/parallel_cluster/bin/start_cluster.sh
+++ b/parallel_cluster/bin/start_cluster.sh
@@ -6,6 +6,13 @@ echo_stderr(){
     echo "${@}" 1>&2
 }
 
+get_local_ip(){
+  # Returns the local ip address
+  local local_ip
+  local_ip=$(dig +short myip.opendns.com @resolver1.opendns.com)
+  echo "${local_ip}"
+}
+
 display_help() {
     # Display help message then exit
     echo_stderr "Usage: $0 NAME_OF_YOUR_CLUSTER "
@@ -27,6 +34,26 @@ check_pcluster_version() {
   # Used to ensure that pcluster is in our PATH
   pcluster version >/dev/null 2>&1
   return "$?"
+}
+
+append_ip_to_extra_parameters() {
+  local extra_parameters
+  extra_parameters="$(echo "$1" | {
+                     # Add AccessFrom attribute - output in raw format
+                     jq ".AccessFrom = \"$(get_local_ip)/0\"" \
+                       --compact-output
+                    } | {
+                      # Compact output is too compact
+                      # Add space in between colons
+                      sed 's/:/: /g'
+                    } | {
+                      # In between commas also helps readability
+                      sed 's/,/, /g'
+                    } | {
+                      # We need these quotes to be escaped
+                      sed 's/"/\\\"/g'
+                    })"
+  echo "${extra_parameters}"
 }
 
 if ! check_pcluster_version; then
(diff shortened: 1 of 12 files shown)


alexiswl:
Not sure this `AccessFrom` is working though.

It's present here as a parameter of the Cloud Formation Stack:
image

But contradicts itself here in the security group settings:
image

parallel_cluster/bin/start_cluster.sh
```diff
@@ -6,6 +6,13 @@ echo_stderr(){
     echo "${@}" 1>&2
 }
 
+get_local_ip(){
+  # Returns the local ip address
+  local local_ip
+  local_ip=$(dig +short myip.opendns.com @resolver1.opendns.com)
```
brainstorm, line 12: Heh, neat trick, I would have used `ip show` or similar (which don't generate an external request) but this works too :)

parallel_cluster/conf/config
```diff
@@ -45,8 +45,8 @@ initial_queue_size    = 1
 # conda has been initialised to the ec2-user
 # See the ami readme for recreation of the custom_ami
 custom_ami            = ami-0309a6a138b994d22
-pre_install           = s3://umccr-temp-dev/Alexis_parallel_cluster_test/bootstrap/pre_install.sh
```
brainstorm, line 48: Much better now, thanks

parallel_cluster/bin/start_cluster.sh
```diff
@@ -117,40 +155,35 @@ fi
...
+if [[ -z "${instance_id}" ]]; then
+  echo_stderr "Could not retrieve instance ID"
+else
+  echo_stderr "Got instance_id ${instance_id}"
+  echo_stderr "Log into your cluster with \"ssm ${instance_id}\""
+fi
+
+# FIXME: control error codes better, avoiding counterintuitive ones: i.e authed within a different account:
+# ERROR: The configuration parameter 'vpc_id' generated the following errors:
+# The vpc ID 'vpc-7d2b2e1a' does not exist
```
brainstorm, line 188: Where does that copy-pasted VPC error come from? :-S

Also, the code above makes it seem like that error management is solved?

brainstorm (approved):

parallel_cluster/bin/start_cluster.sh
alexiswl: If a dev user runs through the Tothill account, I don't have access to that VPC.

alexiswl: I think this is still on the TODO list, or maybe I write something up in the troubleshooting section of the Readme.
I just started the download of the gridss hg38 refdata into an efs system and regretting not prioritising that as a morning task so now looking for things to do...

parallel_cluster/bin/start_cluster.sh
alexiswl: Are you able to replicate `ip show` for getting external IP? I can't seem to find anything on line for it and this seemed to be the shortest solution.

brainstorm: Ah, you are right, what I was thinking does not work with NAT or gateways in general, routes nor ARP caches do not expose those either... (conntracking might)... It's fine this way! ;)

alexiswl:
> Not sure this `AccessFrom` is working though.
> 
> It's present here as a parameter of the Cloud Formation Stack:
> image
> 
> But contradicts itself here in the security group settings:
> image

This has been resolved in 3e95d382be64fa2f157f462014dc2f435f5d89a4with @victorskl suggestion

PR #73, opened 2020-09-09
Added networking stack main vpc security groups

victorskl:
* Added outbound only sg
* Added outbound and uom inbound only to ssh sg

### DIFF
--- a/terraform/stacks/networking/main.tf
+++ b/terraform/stacks/networking/main.tf
@@ -37,7 +37,7 @@ module "main_vpc" {
   private_subnets  = ["10.2.20.0/23", "10.2.22.0/23", "10.2.24.0/23"]
   database_subnets = ["10.2.40.0/23", "10.2.42.0/23", "10.2.44.0/23"]
 
-  // Single NAT Gateway Scenario
+  # Single NAT Gateway Scenario
   enable_nat_gateway     = true
   single_nat_gateway     = true
   one_nat_gateway_per_az = false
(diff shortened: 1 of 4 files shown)


terraform/stacks/networking/main.tf
```diff
@@ -103,3 +103,52 @@ module "main_vpc" {
...
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags = {
+    Name        = "outbound"
```
reisingerf, line 120: I'd use outbound_only, just to be clear

terraform/stacks/networking/main.tf
```diff
@@ -103,3 +103,52 @@ module "main_vpc" {
...
+
+  ingress {
+    from_port   = 22
+    to_port     = 22
+    protocol    = "tcp"
+    cidr_blocks = ["10.1.0.0/20"]
+  }
+
+  tags = {
+    Name        = "uom"
```
reisingerf, line 148: I'd use ssh_from_uom, again to be clearer

terraform/stacks/networking/main.tf
victorskl: Amended

terraform/stacks/networking/main.tf
victorskl: Amended

PR #91, opened 2020-11-24
htsget api gw

victorskl:
- Deploy htsget server based on OpenAPI definition from samtools hts-specs repo upstream
- Implemented htsget service deployment using ref impl


cdk/apps/htsget/requirements.txt
```diff
@@ -0,0 +1,11 @@
+aws-cdk.core
```
reisingerf, line 1: Is there a particular reason why we are placing the dependencies here?
In other apps we placed them in setup.py following the CDK template. Is that no longer recommended?

reisingerf (approved):
Nice work!

cdk/apps/htsget/requirements.txt
victorskl: No particular reason. Happy to move it to setup tool in next, yes.

PR #98, opened 2020-12-14
Refactor SNS/SQS to accomodate >1 consumers

brainstorm:
This is relevant for the companion PR https://github.com/umccr/data-portal-apis/pull/136.


brainstorm:
Not relevant anymore, using one "gateway" lambda dispatch strategy.

PR #102, opened 2021-02-03
Portal TF: Added workflow automation param store config sub-stack

victorskl:


terraform/stacks/umccr_data_portal/workflow/.terraform.lock.hcl
```diff
@@ -0,0 +1,20 @@
+# This file is maintained automatically by "terraform init".
```
reisingerf, line 1: Any particular reason why this file should e kept in GH?

terraform/stacks/umccr_data_portal/workflow/main.tf
```diff
@@ -0,0 +1,231 @@
...
+      source = "hashicorp/aws"
+    }
+  }
+}
+
+provider "aws" {
+  region = "ap-southeast-2"
+}
+
+variable "iap_jwt_token" {}
```
reisingerf, line 23: While I am happy to have all other (non-sensitive) parameters defined in the TF stack, I am not so sure about this one.
There are several reasons:
- most importantly this would make the deployment machine the keeper of the secret, which it should not be
- if I don't have the (current/latest) token locally, I'd need to pull it from somewhere first, right?
- there is a chance that I accidentally push a local token and overwrite the actual/latest production token
- it complicated this stack unnecessarily

Happy to discuss that, especially since we have quite a mixed use of ParameterStore. Maybe we should keep all non-sensitive configuration parameters in ParameterStore (and have that controlled by TF) and keep real secrets in SecretsManager (which is then manually maintained by the admins).

@brainstorm @alexiswl Opinions? Perhaps a topic to quickly discuss in the next DevOps meeting?

reisingerf (changes requested):

terraform/stacks/umccr_data_portal/workflow/.terraform.lock.hcl
victorskl: Yep; according their doco for this deps lock file; similar to `package-lock.json` or `yarn.lock` thing, I reckon.

https://www.terraform.io/docs/language/dependency-lock.html

> You should include this file in your version control repository [...]

terraform/stacks/umccr_data_portal/workflow/main.tf
victorskl: Yep; happy to go manual for token.

brainstorm: Yup, locks in as docs state, tokens out.

reisingerf (approved):

terraform/stacks/umccr_data_portal/workflow/main.tf
alexiswl: A little new to this, is there an issue in keeping it as an encrypted ssm parameter?

reisingerf: No, the main issue is not keeping it in ParameterStore vs SecretsManager, but more where the secret comes from initially (local vs remote).
With the first version it is pushed into ParameterStore from the (local) deployment machine. Now the stack assumes it is already (securely) in place on AWS side.

victorskl: Encrypted SSM parameter is ok, I suppose. Just that we don't wish to mange this through terraform -- but just let it simply directly manage it through Console UI or CLI or as such. The key reason being is that terraform still store value in its own state file of key-value pair.

reisingerf: The secret is stored in plain text in the terraform state in both cases, so the (local) deployment machine has access to it anyway. (solving that is a completely different beast)
However, in the revised/current stack the secret does **not originate** from the client machine (and I believe is not stored in plain text). So a little more secure, but it completely eliminates the issue that the client instance requires the secret as prerequisite for the stack to work, and that different client machines can have different/conflicting secrets.

PR #103, opened 2021-02-05
Pull token for GH provider from SSM instead of local envvar

reisingerf:
Remove the need to have a local GH token and pull it instead from SSM ParameterStore.


victorskl (approved):
Wow! Neat! 👍

brainstorm (approved):
Cool!

PR #136, opened 2021-06-08
Mapped production project and workgroup to dev t/n workflow

alexiswl:
Moved over to dev workflow to project based through the cwl-ica repo. Took the fastest approach.

A couple of issues here:

1. Prod version not GIT version controlled, I cannot create a workflow in the prod project because I do not have permissions.  
2. I am unsure if this workflow should be in the workgroup or project. Are we launching all future runs through projects?

For now I have updated the dev project workflow that @skanwal ported over to the cwl-ica repo to also be part of the dev and prod workgroups and dev and prod projects (for workflow and workflow versions) through updating the acl parameter on the cli.

A couple of solution options to the issues raised above:

1. I clone the definition / re-pack on cwl-ica and then manually create a  workflow version for the production workgroup, then update the acl so that the workflow and the workflow version can be seen by the production project.  

2. We create a 'scoped' token with the scopes [here](https://github.com/umccr/cwl-ica/blob/beta-release/src/utils/globals.py#L12-L54). The token would not have write (or even read access) to production data.  I can then manage the workflows through the cwl-ica repo.

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -152,7 +152,7 @@ locals {
 
   tumor_normal_wfl_version = {
     dev  = "3.7.5"
-    prod = "3.7.5"
+    prod = "3.7.5--5a56dd"
   }
 
   tumor_normal_input = {
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -146,13 +146,13 @@ locals {
   }
 
   tumor_normal_wfl_id = {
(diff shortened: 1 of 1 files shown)


reisingerf:
Ideally we'd only run in `project` context, yes.
Happy to use a token, but I am not sure if scopes would even apply in a project context (in theory they should **not**). We may have to go the two step approach (create - authorise).
Happy to try/discuss options.

victorskl (commented):
Conclusion is:

1. Short-term, for this PR, @alexiswl will add git commit short ID to this production workflow version
2. Long-term solution is to use dedicated ICA Project context only for workflow deployment purpose (not for running workflows)

Thanks guys. /cc @alexiswl @reisingerf @skanwal

victorskl (approved):
LGTM! Going to merge and apply both DEV and PROD. /FYI

PR #138, opened 2021-06-11
Rolled back from 'ica' namespace to 'iap' namespace

alexiswl:
Sorry, will get my terraform config sorted soon enough. This one still needs to be applied.


victorskl (approved):
No worries. Flo, I take this. Apply next in tick.

PR #151, opened 2021-07-06
Add cttso workflow to ssm

alexiswl:
Required by https://github.com/umccr/data-portal-apis/pull/232 for automation of ctTSO analysis on portal

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -264,17 +264,17 @@ locals {
   }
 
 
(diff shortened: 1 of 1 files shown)


victorskl (approved):
Just that, does workflow type naming finalise with `dragen_cttso`? 

And that would be sufficient enough to differentiate as discussed in https://github.com/umccr/data-portal-apis/issues/226

LGTM, otherwise!

alexiswl:
Ah good point, will move it from `dragen_ct_tso` to `dragen_tso_ctdna`

reisingerf (approved):
LGTM

PR #157, opened 2021-07-26
Slack CDK: use individual rules for Batch queues

scwatts:
Replace the catch-all CloudWatch event rule for Batch job status change
with individual rules for each Batch job queue. This allows Slack
notifications for specific job queues to be enabled or disabled.

### DIFF
--- a/cdk/apps/slack/stacks/slack_lambda.py
+++ b/cdk/apps/slack/stacks/slack_lambda.py
@@ -85,7 +85,11 @@ class BatchLambdaStack(core.Stack):
             'WtsReport':      {'name': 'wts_report_batch_queue_dev',             'enabled': True},
         }
         for job_queue_id, job_queue_config in batch_job_queues.items():
-            rule = _events.Rule(
+            aws_account_id = kwargs['env']['account']
+            aws_region = kwargs['env']['region']
+            job_queue_name = job_queue_config['name']
+            job_queue_arn = f'arn:aws:batch:{aws_region}:{aws_account_id}:job-queue/{job_queue_name}'
+            _events.Rule(
                 self,
                 f'BatchEventToSlackLambda{job_queue_id}',
                 enabled=job_queue_config['enabled'],
@@ -96,7 +100,7 @@ class BatchLambdaStack(core.Stack):
                             'SUCCEEDED',
                             'RUNNABLE',
                         ],
-                        'jobQueue': job_queue_config['name'],
+                        'jobQueue': [job_queue_arn],
                     },
                     detail_type=['Batch Job State Change'],
                     source=['aws.batch'],
--- a/cdk/apps/slack/stacks/slack_lambda.py
+++ b/cdk/apps/slack/stacks/slack_lambda.py
@@ -85,10 +85,8 @@ class BatchLambdaStack(core.Stack):
             'WtsReport':      {'name': 'wts_report_batch_queue_dev',             'enabled': True},
         }
         for job_queue_id, job_queue_config in batch_job_queues.items():
-            aws_account_id = kwargs['env']['account']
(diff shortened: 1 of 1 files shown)


cdk/apps/slack/stacks/slack_lambda.py
```diff
@@ -77,17 +77,30 @@ def __init__(self, scope: core.Construct, id: str, slack_channel: str, **kwargs)
...
+                f'BatchEventToSlackLambda{job_queue_id}',
+                enabled=job_queue_config['enabled'],
+                event_pattern=_events.EventPattern(
+                    detail={
+                        'status': [
+                            'FAILED',
+                            'SUCCEEDED',
+                            'RUNNABLE',
+                        ],
+                        'jobQueue': job_queue_config['name'],
```
scwatts, line 99: Thinking further about this, the value of `jobQueue` will likely need to be set as the job queue ARN rather than the job queue name. The CDK JobQueue class has a [method](https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_batch/JobQueue.html#aws_cdk.aws_batch.JobQueue.from_job_queue_arn) to look up a defined job queue but the lookup itself requires the ARN.

Is it safe to construct ARNs where the resource id is known as shown [here](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax)? If so, this could be alternative.

reisingerf: Yes, I think that has to be an ARN and yes, it should be safe to construct the ARN. The only variable parts are the account ID (which you can get from CDK) and the queue name (which you already have there).

reisingerf (approved):
Looks good! Happy for you to merge and try (perhaps with the ARN change).

scwatts:
The stack successfully deployed on dev and appears to be working. Should we consider limiting rules to accounts that have the corresponding job queues? The prod account doesn't currently have a nextflow Batch job queue and so doesn't necessarily need the event rule.

I'm not sure this is a good idea as the solution then becomes slightly less generalised and adds some maintenance overhead. Regardless, this would be easy to implement - associate job queues with specific accounts in the `batch_job_queues` dictionary and operate on entries appropriately within the loop.

cdk/apps/slack/stacks/slack_lambda.py
```diff
@@ -77,17 +77,34 @@ def __init__(self, scope: core.Construct, id: str, slack_channel: str, **kwargs)
...
+        batch_job_queues = {
+            'AGHA':           {'name': 'agha-job-queue',                         'enabled': True},
+            'Umccrise':       {'name': 'cdk-umccrise_job_queue',                 'enabled': True},
+            'UmccriseDragen': {'name': 'cdk-umccrise_job_queue-dragen-testing',  'enabled': False},
+            'Nextflow':       {'name': 'nextflow-job-queue',                     'enabled': False},
+            'WtsReport':      {'name': 'wts_report_batch_queue_dev',             'enabled': True},
+        }
+        for job_queue_id, job_queue_config in batch_job_queues.items():
+            aws_account_id = kwargs['env']['account']
+            aws_region = kwargs['env']['region']
```
scwatts, line 89: Looking at the CDK docs I wasn't able to see an obvious way to reference the `region` and `account` variables that are passed to the `Stack` constructor:
https://github.com/umccr/infrastructure/blob/cb80fd0f7096b8e76e932d592d4b48d91e020832/cdk/apps/slack/app.py#L20-L25

Using `kwargs` doesn't seem like the most robust approach. Any suggestions?

scwatts:
The latest commit changes how AWS region and account values are accessed, as discussed. Ready to merge on my end!

reisingerf:
Looks good. Feel free to merge.

PR #175, opened 2021-10-12
Ontoserver ALB -> Fargate Docker

andrewpatto:


andrewpatto:
Brand new stack but if anyone wants to review for possible improvements go for it

PR #183, opened 2021-11-11
Initial commit of cttso-ica-to-pieriandx-cdk

alexiswl:
TODO List:

- [x] What to do with passwords / tokens?
  * ICA_ACCESS_TOKEN
  * PIERIANDX_USER_EMAIL
  * PIERIANDX_USER_PASSWORD
- [x] Also add in dev ssm parameters
  * ICA_BASE_URL
  * PIERIANDX_BASE_URL
   * PIERIANDX_INSTITUTION
   * PIERIANDX_AWS_REGION
   * PIERIANDX_AWS_S3_PREFIX
- [x] AWS Code Pipeline for Dockerbuild, currently manual build on quay.io
- [x] Fix complaints in PyCharm re aws-cdk v2.0 not having quite the same 'batch' methods 
- [x] Test / Deploy in dev
- [x] Add in prod ssm parameters
- [ ] Test / Deploy in prod


victorskl (commented):
> re: ICA_ACCESS_TOKEN & ICA_BASE_URL

If the processing impl inside Docker is using `libica` then latest [release 1.1.0](https://github.com/umccr-illumina/libica/tags) added `libica.app` [App package](https://github.com/umccr-illumina/libica#using-app-package). It contains configuration object builder which streamline setting this two settings and, you can instead load `ICA_ACCESS_TOKEN` from Secret Manager; e.g.

```python
gds_config = configuration(lib=libgds, secret_name="IcaSecretsWorkflow")
```

And, `libica.app.gds` also contains [those methods](https://github.com/umccr-illumina/libica/blob/main/libica/app/gds.py) we implemented together there from Portal, so have a look...

> re: PIERIANDX_*

Or anything in general; that need load from Parameter Store or Secret Manager, you could [utilise `libumccr`](https://github.com/umccr/libumccr) -- (_which is how `libica` use to load from secret manager by the way_) -- like so:

```python
from libumccr.aws import libsm, libssm

pdx_pass = libsm.get_secret(secret_name="PierianDxUserPassword")

pdx_base_url = libssm.get_ssm_param(name="/pdx/base_url")
```

HTH

alexiswl:
Thanks @victorskl, I'll have a look through these modules tomorrow

alexiswl:
> you can instead load ICA_ACCESS_TOKEN from Secret Manager; e.g.

But I would need aws access to do this right? Is this something that the lambda launching the container can solve first and pass the ICA_ACCESS_TOKEN to the container?

 > Or anything in general; that need load from Parameter Store or Secret Manager, you could utilise libumccr -- (which is how libica use to load from secret manager by the way) -- like so:

This looks quite helpful! Again I think we can use this for the ICA_ACCESS_TOKEN too

victorskl:
> But I would need aws access to do this right? 

Yes

> Is this something that the lambda launching the container can solve first and pass the ICA_ACCESS_TOKEN to the container?

That's one way -- you can pass around JWT as environment variables, yes. Or better yet, just attach the the AWS execution role (ECS) and/or Ec2 instance role (Batch) which is having policy to access Secret Manager / Param Store. Then, let SDK library (be it boto or libumccr or libica, etc...) get it from there, quite natively. And avoid passing around token.

alexiswl:
> That's one way -- you can pass around JWT as environment variables, yes. Or better yet, just attach the the AWS execution role (ECS) and/or Ec2 instance role (Batch) which is having policy to access Secret Manager / Param Store. Then, let SDK library (be it boto or libumccr or libica, etc...) get it from there, quite natively. And avoid passing around token.

Where are the roles set up by CDK for that actual process that is running the wrapper script inside the docker container? 
I've just used the aws cli to collect the secrets before launching the python script.  I think I can just pass the rest of the env vars in through ssm parameters.

alexiswl:
I don't know if this is going to make it to the Sokoke tag? The RedCap metadata might not be there by that point and instead this is going to need to be a manual trigger of the lambda function in order to trigger the pipeline.  Other than deployment to prod, is there anything else missing in this one?

alexiswl:
Is it okay for me to deploy this to prod?

victorskl:
Yup; let play by ear. Sokoke release date can/may be adjusted, yes.

victorskl:
> Is it okay for me to deploy this to prod?

Probably ok. But, let me have a peek and catch up with this branch. Probably chat together with bulk reprocess samples.

victorskl (approved):
LGTM! Scope back into Ragdoll release!

alexiswl:
Now in https://github.com/umccr/cttso-ica-to-pieriandx/blob/main/deploy

PR #189, opened 2021-11-25
Updated dragen somatic production version

alexiswl:
* New version collects the normal_bam_out attribute correctly with new name change to normal bam file.


victorskl (approved):
LGTM! ditto Ragdoll release!

PR #193, opened 2021-12-13
Added env variable to disable log4j problem

andrewpatto:
Added availabilty zones

### DIFF
--- a/cdk/apps/ontoserver/ontoserver_stack.py
+++ b/cdk/apps/ontoserver/ontoserver_stack.py
@@ -80,9 +80,11 @@ class OntoserverStack(core.Stack):
                 "Stack": "networking",
             },
         )
+
+        # if we were to want high availability - we would want the instances spread over our zones
         private_subnets = ec2.SubnetSelection(
             subnet_type=ec2.SubnetType.PRIVATE,
-            availability_zones=["ap-southeast-2a"],
+            availability_zones=["ap-southeast-2a", "ap-southeast-2b", "ap-southeast-2c"],
         )
 
         sg_elb = ec2.SecurityGroup(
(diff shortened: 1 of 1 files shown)


andrewpatto:
Also changed fargate security zone to disable outgoing traffic as onto server does not make any HTTP calls to anywhere (as far as I can tell)

andrewpatto:
I haven't actually tested how well this env variable works - I saw it in some of the discussions re: the log4j problem. Am hoping there will be an actual update of ontoserver soon that uses a fixed log4j

reisingerf (approved):
LGTM

brainstorm:
Is ontoserver somewhere publicly available? Tempted to just send them a PR bumping it there on their `pom.xml` or `build.gradle` and be done, but seems like this env var is an [approved mitigation](https://blog.qualys.com/vulnerabilities-threat-research/2021/12/10/apache-log4j2-zero-day-exploited-in-the-wild-log4shell) as you mentioned, so LGTM so far... [not too tricky to exploit/test](https://github.com/jas502n/Log4j2-CVE-2021-44228), btw... also there is middleware to detect and filter exploits too: https://github.com/OllieJC/aws-log4j-mitigations

brainstorm (approved):
LGTM

brainstorm:
Uh-oh, apparently there are updates that state that `LOG4J_FORMAT_MSG_NO_LOOKUPS` is not enough? (official source: https://logging.apache.org/log4j/2.x/security.html)

>Other insufficient mitigation measures are: setting system property log4j2.formatMsgNoLookups or environment variable LOG4J_FORMAT_MSG_NO_LOOKUPS to true for releases >= 2.10, or modifying the logging configuration to disable message lookups with %m{nolookups}, %msg{nolookups} or %message{nolookups} for releases >= 2.7 and <= 2.14.1.

This should really be fixed upstream... is the code for this thing opensource?

andrewpatto:
its not open source.. turns out they don't use log4j (in the version we currently use).

The security group changes here are worthwhile anyhow..

PR #194, opened 2021-12-16
TF: Add stack to manage UoM administration access

reisingerf:


brainstorm (approved):
Not too keen on yet another IAM user (thinking about AWS SecurityHub rightfully complaining) but the alternatives are worse/suboptimal. LGTM & thanks Florian!

reisingerf:
SecurityHub should not complain (I added the user to an IAM group), but I agree, not ideal.
I thought about onboarding Jen on our Google Workspace, but that is similarly sub-optimal.
Let's hope AWS can come up with a better solution.

PR #201, opened 2022-02-17
Update production version of umccrise on ica from 2.0.0 to 2.0.1--3.9.3

alexiswl:
- [x] Waiting on merge of #200

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -329,7 +329,7 @@ locals {
 
   umccrise_wfl_version = {
     dev = "2.0.0--3.9.3"
-    prod = "2.0.0--3.9.3--5d22f74"
+    prod = "2.0.1--3.9.3--092ccf4"
   }
 
   umccrise_wfl_input = {


scwatts (approved):

victorskl:
Linking related upstreams:
- https://github.com/umccr/umccrise/pull/83
- https://github.com/umccr/cwl-ica/pull/78

victorskl (approved):
I reckon, we are ok to backport this into current Sokoke release (i.e. going hot into PROD); since it contains a fix. Otherwise, we can schedule with Korat (couple more weeks, eta). Pls advice workflow-wise. Happy to apply terraform accordingly.

alexiswl:
@scwatts I think hotfix is the way to go here right? We'd want to rerun those samples that failed umccrise in prod right?

ohofmann:
Don't think we have to re-run prod other than to show that it worked.

scwatts:
I'd prefer backporting into the Sokoke release to be sure that the problem identified in prod is resolved rather than waiting a couple of weeks. I don't think we need to rerun all of those samples though - Umccrise is running on ICA as essentially a proof of concept at the moment to uncover these types of problems. Maybe rerunning one or two of the previously failed samples would be a good demonstration though since I'm unsure when more top-up samples would be coming through.

victorskl:
Thanks all for chime in. @alexiswl Please kindly merge and, I will then apply it to PROD; and run the most recent failed `SBJ01560` to determine it works.

PR #206, opened 2022-03-08
Added ability to build a v2 token.

alexiswl:
Would be good to be able to access this via instances or other places we login rather than copying tokens out to clipboard etc.

### DIFF
--- a/cdk/apps/ica_credentials/app.py
+++ b/cdk/apps/ica_credentials/app.py
@@ -10,6 +10,7 @@ CDK_APP_NAME = "ica-credentials"
 CDK_APP_PYTHON_VERSION = "3.8"
 
 ICA_BASE_URL = "https://aps2.platform.illumina.com"
+ICAV2_BASE_URL = "https://ica.illumina.com"
 
 SLACK_HOST_SSM_NAME = "/slack/webhook/host"
 SLACK_WEBHOOK_SSM_NAME = "/slack/webhook/id"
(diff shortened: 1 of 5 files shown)


cdk/apps/ica_credentials/app.py
```diff
@@ -47,7 +48,25 @@
...
+# V2 (single token)
+IcaCredentialsDeployment(
+    app,
+    f"{CDK_APP_NAME}-dev-v2",
+    None,  # Token does not require project context in v2
+    None,  # Token does not require additional project list in v2
+    ICAV2_BASE_URL,
+    SLACK_HOST_SSM_NAME,
+    SLACK_WEBHOOK_SSM_NAME,
+    env=cdk.Environment(
```
andrewpatto, line 66: So is there going to be no difference between our ica v2 settings between *our* prod and dev?

andrewpatto (approved):

cdk/apps/ica_credentials/app.py
alexiswl: I don't know. For now probably not, and we don't have a prod account? I left a comment regarding the `encoded_params` since we can specify a tenant in V2 but we currently only have the one.  There may be potential to restrict the scope of the api key? So we could have one api key that is part of workgroup 'dev-admin' which has admin access to the dev project but only read access to the prod project? 

@reisingerf @victorskl thoughts?

victorskl: The proper ICA dev/prod isolation should be at tenant level, TBH. Different projects within the same ICA tenant still share the compute environment. At project level, we only can achieve "data" isolation.

Likewise in v1 and GDS; data will be still scoped into Project in v2 i.e. `icav2 projectdata list`. It is just that, project scoped for data is explicit in v1 (i.e. we assign who can write/read to GDS volume) but implicit in v2 (creating a project implicitly get data volume). So, there's still a dedicated project needed for the production pipeline, I suppose.

victorskl (approved):

reisingerf (approved):
LGTM for now.
We'll see how the permission system works out in v2.

victorskl (commented):
FYI
- Going to merge this in tick next; to get us started testing v2 token in DEV. 
- If it goes well, schedule to apply changes to PROD when closing [Chartreux](https://github.com/umccr/infrastructure/milestone/6?closed=1) release.

PR #198, opened 2022-02-13
Updated bclconvert workflow

alexiswl:
* Override cycles matching '\w+;\w+;[nN]\d+;[nN]\d+' now auto-removes adapterread2 from samplesheet settings

---

- [ ] @victorskl to follow up apply to PROD when closing Korat milestone.

PR #207, opened 2022-03-09
Updated WTS workflow inputs

skanwal:
- copied validation to `production` volume (courtesy Alexis)
- updated data in `development`

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -268,7 +268,11 @@ locals {
               "location": "gds://development/validation_data/wts/SBJ00028_PRJ190190_L1900423/analysis/dragen_wts/2021-12-01--3.9.3/SBJ00028"
           }
       ],
-      "cl_config": "{\"sample_names_replace\":{\"L1900440\":\"Ref_1\",\"L1900877\":\"Ref_2\",\"L1900920\":\"Ref_3\",\"L1901023\":\"Ref_4\",\"L1901028\":\"Ref_5\",\"L1900423\":\"Ref_6\"}}"
+      "cl_config": "{\"sample_names_replace\":{\"L1900440\":\"Ref_1\",\"L1900877\":\"Ref_2\",\"L1900920\":\"Ref_3\",\"L1901023\":\"Ref_4\",\"L1901028\":\"Ref_5\",\"L1900423\":\"Ref_6\"}}",
+      "sites_somalier": {
+            "class": "File",
+            "location": "gds://development/reference-data/somalier/sites.hg38.rna.vcf.gz"
+        }
     }
     EOT
     prod = <<-EOT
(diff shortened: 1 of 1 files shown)


terraform/stacks/umccr_data_portal/workflow/main.tf
```diff
@@ -62,7 +62,7 @@ locals {
 
   bcl_convert_wfl_version = {
     dev  = "3.7.5"
-    prod = "3.7.5--fade666"
+    prod = "3.7.5--f1e67a3"
```
alexiswl, line 65: This looks like it has regressed from fade666 (which was on 10 Feb 2022) to f1e67a3 (which was on 7 June 2021)

alexiswl, line 65: This looks like it has regressed from fade666 (which was on 10 Feb 2022) to f1e67a3 (which was on 7 June 2021)

alexiswl (changes requested):
Hey @skanwal looks like this has changed the workflow versions as well. I rebased the branch onto [update-dragen-workflow-versions-prod](https://github.com/umccr/infrastructure/tree/update-dragen-workflow-versions-prod) but it looks like the bcl_convert_wfl_version on this branch is still the old version (f1e67a3) instead of (fade666) or (7afeacf).  

I can rebase, soft-reset, recommit and force-push if need be which then this PR contains JUST the changes to the workflow inputs if that's helpful?

alexiswl:
Also worth considering, in the BCLConvert workflow we use the `cl_config` command to set the genome size to `hg38_genome` see https://github.com/umccr/cwl-ica/blob/main/workflows/bcl-conversion/3.7.5/bcl-conversion__3.7.5.cwl#L218-L221 for more details.  

Is it worth placing this `cl_config` in the arguments of the dragen transcriptome pipeline where you can then set the names of the reference samples, rather than pointing to a file? If a file on GDS changes or is deleted, it can be hard to know what was run, whereas this way we can always look back on a workflow run to see what the configuration was.

skanwal:
> Is it worth placing this cl_config in the arguments of the dragen transcriptome pipeline
Basically you're suggesting to create the `replace_names` tsv with in the workflow def?

terraform/stacks/umccr_data_portal/workflow/main.tf
skanwal: This is what was set in your PR in the `main.tf` file. I basically took your changes and added WTS workflow input changes.

terraform/stacks/umccr_data_portal/workflow/main.tf
```diff
@@ -209,7 +209,7 @@ locals {
 
   wts_tumor_only_wfl_version = {
     dev  = "3.9.3"
-    prod = "3.9.3--9961e75"
+    prod = "3.9.3--7ba55e9"
```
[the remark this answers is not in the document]
skanwal: Okay from my side if you want to rebase it.

terraform/stacks/umccr_data_portal/workflow/main.tf
alexiswl: Ah okay, so it would have been the old version because the dragen workflow prod branch didn't update the bcl convert, that was done by https://github.com/umccr/infrastructure/pull/198, and the dragen workflow prod branch stemmed from master before that was merged in.

alexiswl: But by setting it here it's seen as a change in the file commit which is upstream of #198.

skanwal:
First need to merge https://github.com/umccr/cwl-ica/pull/86/files

PR #209, opened 2022-03-13
Updated somalier and reference file

alexiswl:
## Updates

* Change somalier file for alignment qc workflow to rna file 
* Change reference data to fasta in root volume

* gds volumes umccr-refdata-dev and umccr-refdata-prod are no longer used for automated workflows

## TODO 
- [x] Compare md5sum of reference fasta files (items in umccr-refdata-x do NOT have an eTag)
  * Both md5sums are evaluated to `64b32de2fc934679c16e83a2bc072064`

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -126,11 +126,11 @@ locals {
       "fastq_list_rows": null,
       "sites_somalier": {
         "class": "File",
-        "location": "gds://umccr-refdata-dev/somalier/sites.hg38.vcf.gz"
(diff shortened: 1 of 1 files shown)


victorskl (approved):
Yep! can confirm reference genome matches. 

For somalier sites info; this means that they were the same between WGS QC alignment and WTS transcriptome?

alexiswl:
> For somalier sites info; this means that they were the same between WGS QC alignment and WTS transcriptome?

Yes that's correct

victorskl (commented):
FYI, if no more comment, I gonna to merge it in tick, next few minutes. Thanks.

PR #210, opened 2022-03-14
Reference data directory should be class Directory not File

alexiswl:
Reference data directory should be of class 'Directory' not 'File'


victorskl (approved):
Yep; thanks for the quick patch. Me also figure on the first pass of validation Run.

PR #213, opened 2022-03-15
Added prod workflow version for rnasum

alexiswl:


victorskl (approved):
LGTM! Merging it now. I will apply to PROD when releasing Korat.

PR #215, opened 2022-03-24
Hardcode dataset for RNASum

alexiswl:


victorskl (approved):
Thanks @alexiswl I can backport this to Korat!

PR #220, opened 2022-04-20
Pin Umccrise reference data version to 1.0.11

scwatts:
* the DVC pull command for Umccrise reference data operates on the .dvc file at HEAD on master
* upcoming changes to Umccrise reference data may impact the ability to run Umccrise 2.0.1
* this change pins Umccrise reference data pulled by DVC to 1.0.11


reisingerf (approved):

victorskl (approved):
Ah correct! Checkout to a tag version or a (released) branch is recommended way to use dvc versioned data before pull. HEAD may be ok, only if, "a merge to the main branch means a release" implicit.

LGTM!

PR #221, opened 2022-04-27
Update s3 policy for `mm` iam group

williamputraintan:


reisingerf (commented):
No need to change the bucket policy.
The access is controlled via the policies attached to the IAM group (that policy had a typo in the resource condition which I fixed already).

The bucket policy for `nswhp` is because they want a direct AWS to AWS account access (without using any of our IAM users/roles), so we've allowed direct access to the bucket from their account via bucket policy.

reisingerf:
Thanks for checking this, but I think we can delete this PR.

PR #222, opened 2022-04-27
Updated rnasum and wts prod workflow versions to fix arriba output issue

alexiswl:


victorskl (approved):
LGTM. Arriba fixes. This will be backporting to the current Korat release and, will push straight to PROD. FYI.

PR #223, opened 2022-04-27
bclconvert-workflow version update

alexiswl:
* Remove MinimumAdapterOverlap if AdapterRead1 and AdapterRead2 are not present


victorskl (approved):
LGTM. 10X adapter failed-safe check fixes. This will be backporting to the current Korat release and, will push straight to PROD. FYI.

PR #226, opened 2022-05-12
Portal TF: Added Portal UI CodeBuild env vars for GPL links

victorskl:
* i.e. Service discovery for GPL Function URLs from SSM parameter

Related
* https://github.com/umccr/gridss-purple-linx-nf/pull/20
* https://github.com/umccr/data-portal-apis/issues/431


victorskl (commented):
At some point, we will need better service discovery, if this become mainstream pattern... FYI

PR #235, opened 2022-06-23
Bastion TF: Fixed properly remove packer user

victorskl:
Flo; I reckon `packer` user must be deleted through Console at some point (probably during SecurityHub sweep). Anyway, I just re-sync it with this terraform state and, reflect it now.


reisingerf (approved):

victorskl:
Ah, you are up, early, again!

PR #237, opened 2022-07-12
Portal TF: Added Athena setup for Portal Data Catalog

victorskl:
* Integrated through Athena Query Federation SAM application i.e. possibility
  with federated queries across all data pipelines output as data sources mash-up.
  See https://docs.aws.amazon.com/athena/latest/ug/writing-federated-queries.html

### DIFF
--- a/terraform/stacks/umccr_data_portal/athena.tf
+++ b/terraform/stacks/umccr_data_portal/athena.tf
@@ -1,0 +1,71 @@
+locals {
+  secret_name_prefix = "data_portal/rds/master"
+}
+
+resource "aws_serverlessapplicationrepository_cloudformation_stack" "athena_jdbc_connector" {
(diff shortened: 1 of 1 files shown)


victorskl:
@reisingerf @brainstorm Please feel free to review when you are able. This setup Athena with terraform. And Athena Query Federation SAM repo for DB connector.

I am thinking similar for AG/DynamoDb, next. But it might need Glue in between. Keep you all posted. FYI.

williamputraintan (approved):

williamputraintan:
I would assume at some point we will need to clear result from S3 regularly?

andrewpatto (approved):

victorskl:
> clear result from S3 regularly?

Good point, Will. Absolutely, yes. I will add life cycle rule on Athena result and spill prefixes.

PR #249, opened 2022-08-23
Update RNAsum workflow versions

skanwal:
- Updated both dev and prod versions.

- Corresponds to https://github.com/umccr/RNAsum/pull/71


victorskl (approved):
Looking good. As usual, if no input changes needed then this is good to go, yes!

alexiswl (approved):

PR #252, opened 2022-09-21
Datalake TF: Added umccr_datalake warehouse stack

victorskl:
* Initial setup with multiqc Glue cralwers for Parquet data lake

### DIFF
--- a/terraform/stacks/umccr_datalake/policies/datalake_role_inline_policy.json
+++ b/terraform/stacks/umccr_datalake/policies/datalake_role_inline_policy.json
@@ -1,0 +1,33 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:GetObject",
+        "s3:PutObject"
+      ],
+      "Resource": [
+        "arn:aws:s3:::${bucket}/*"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "lakeformation:*"
+      ],
+      "Resource": [
+        "*"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "kms:Decrypt"
+      ],
+      "Resource": [
+        "${sse_key_arn}"
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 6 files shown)


terraform/stacks/umccr_datalake/policies/datalake_role_inline_policy.json
```diff
@@ -0,0 +1,33 @@
...
+        "s3:PutObject"
+      ],
+      "Resource": [
+        "arn:aws:s3:::${bucket}/*"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "lakeformation:*"
```
pdiakumis, line 17: Nice, I assume this policy would've helped the other day when I had to try all sorts of lake formation permission combinations to get through these errors: https://aws.amazon.com/premiumsupport/knowledge-center/glue-insufficient-lakeformation-permissions/

pdiakumis (approved):

reisingerf (approved):
Sweet!

terraform/stacks/umccr_datalake/policies/datalake_role_inline_policy.json
victorskl: Good pointer to the article, Pete. TBH; I just copy in this policy from the Glue role that auto setup through Glue Console. Let me read up the article and, see if I can fine tune a bit. Looks like we could/should tune a bit here, yes.

brainstorm (approved):
Cool! So nice to see this rolling ;)

terraform/stacks/umccr_datalake/policies/datalake_role_inline_policy.json
victorskl: Right! Done a bit of reading. And removed LakeFormation permission with https://github.com/umccr/infrastructure/commit/d7bf0859418e9e506d3ef578d2ba84cca4bc2424

* Turn out we do not need LakeFormation permission just yet as we can just use Glue default IAM-based permission.
* We can upgrade to LakeFormation access control model in future when it sees fit for more complex access control use cases.
* For that, see
https://docs.aws.amazon.com/lake-formation/latest/dg/upgrade-glue-lake-formation.html

victorskl:
Thanks all, merging...

PR #253, opened 2022-10-10
Short term fix for rotation ICA

andrewpatto:
Added missing libs to reqs
Changed rotation to 1 day


alexiswl:
The curse of the very fast-paced CDK where one line change becomes 500! Hopefully once this is in a self-mutating pipeline things are a little bit more stable

alexiswl (approved):

PR #257, opened 2022-10-21
umccrise-with-germline-pipeline updates

alexiswl:
## Summary
* Turn on enable_duplicate_marking for umccrise_with_germline pipeline 
* Upgraded umccrise version in dev and prod from 2.1.1 to 2.2.0
* ICA Prod workflow version changes from 2.1.1--3.9.3--e4195a6 to 2.2.0--3.9.3--4e00721

## Related PRs
* https://github.com/umccr/cwl-ica/pull/151
* https://github.com/umccr/cwl-ica/pull/152

---

DevOps note -- before applying to PROD:

* [ ] Victor/Flo to run this changes pass through the whole Pipeline i.e End-to-end (e2e) testing with validation dataset or some PROD grade WGS T/N cases in STG

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/main.tf
+++ b/terraform/stacks/umccr_data_portal/workflow/main.tf
@@ -382,8 +382,8 @@ locals {
   }
 
   umccrise_wfl_version = {
-    dev = "2.1.1--3.9.3"
-    prod = "2.1.1--3.9.3--e4195a6"
+    dev = "2.2.0--3.9.3"
+    prod = "2.2.0--3.9.3--4e00721"
   }
 
   umccrise_wfl_input = {
@@ -394,6 +394,7 @@ locals {
       "output_directory_germline": null,
       "output_directory_umccrise": null,
       "output_file_prefix_germline": null,
+      "enable_duplicate_marking": true,
       "reference_tar_germline": {
         "class": "File",
         "location": "gds://development/reference-data/dragen_hash_tables/v8/hg38/altaware-cnv-anchored/hg38-v8-altaware-cnv-anchored.tar.gz"
@@ -412,6 +413,7 @@ locals {
       "output_directory_germline": null,
       "output_directory_umccrise": null,
       "output_file_prefix_germline": null,
+      "enable_duplicate_marking": true,
       "reference_tar_germline": {
         "class": "File",
         "location": "gds://production/reference-data/dragen_hash_tables/v8/hg38/altaware-cnv-anchored/hg38-v8-altaware-cnv-anchored.tar.gz"


skanwal (approved):
LGTM!

alexiswl:
https://github.com/umccr/cwl-ica/blob/main/.github/workflows/sync-tools_and_sync-workflows.yml#L40-L66 should instead use the .pull_request.head.ref attribute to get the branch name.  

We can also use this to collect the git commit id at .pull_request.head.sha

pdiakumis (approved):
LGTM too, though not sure if @victorskl wanted to merge this now or beginning of November for the portal to pick up.

alexiswl:
Ah good point, we just turn on enable_duplicate_marking in the next release then along with the upgrade to umccrise?

pdiakumis:
Yep, I should have the comparison of the dup vs. non-dup validation samples ready by then.

victorskl (approved):
Looking good, folks! 

Just answer -- typically, I'd flag `pending apply prod` label and, apply it to PROD when release milestone is finalised. Unless  specifically stated, otherwise.

Just wanted to double confirming: we wanted to set  `enable_duplicate_marking` as `true` for most cases, correct?

Also. Just want to capture, say, which kind of event or when would we want to switch this flag to `false`. Just for Pipeline Automation benefit / requirement gathering purpose.

pdiakumis:
- Q: Just wanted to double confirming: we wanted to set enable_duplicate_marking as true for most cases, correct?
  - A: Yes, for the WGS cases. For WTS cases it's a different story.

- Q: Which kind of event or when would we want to switch this flag to false
  - A: WTS cases. I am experimenting on the side to see what impact setting/unsetting this flag has on our WGS validation samples.

victorskl:
Awesome, thanks for explain, Pete!

victorskl (commented):
Due to few changes observed upstream cwl-ica PRs (black box observation); 

- raised the flag [e2e testing required](https://www.google.com/search?q=end-to-end+(E2E)+testing) for the Automated Pipeline
- added Oliver for peer review, if any

Side Note:
Alexis/Sehrish; just heads up, would be good if we still have any other cwl-ica changes are inbound come in time for next release. Then, we can run e2e testing with one sweep run. Happy to coordinate further on next week, if any.

ohofmann (commented):
A question, mostly - why is the `enable_duplicate_marking` flag the responsibility of the portal? Shouldn't this be encoded in the CWL workflows and fixed there? If we have the launcher change workflow parameters at runtime I need to find a way to capture this in the documentation; it also adds another layer of code to review wrt workflows.

reisingerf:
Hm, that would be a shift in our current philosophy/setup?

As I understand, our workflows are basically tools (or aggregations of tools to create a new tool, similar to a lot of docker container use cases) meant to be useful for more than one use case. 
What you are suggesting is fixing every possible parameter in order to enforce a very specific behaviour of the tool? Which would also result in a new version if the tool stays the same and only a parameter changes.

I guess we could do that and create "execution versions" of "tool versions"... I can see that it could help... but... 
I am not sure how much it would help though. I guess there are always inputs that can't be fixed and if those have to be tracked we probably need to take the workflow inputs into account anyway?
Thinking about things like how we decide on demultiplexing settings, T/N pairing, TSO metadata?, ... 

Most of those parameters are now part of the workflow input template (the one that also defines the actual workflow version to use), and that's controlled via versioned code and tracked as part of a portal release. (So could be reviewed, updated and released separately as well, without touching portal or workflows)

Perhaps worth discussing in more detail independently of this PR?

ohofmann:
> Perhaps worth discussing in more detail independently of this PR?

Yep, agreed.

alexiswl:
> Perhaps worth discussing in more detail independently of this PR?

Sorry, I meant to bring this up today at the bioinfo meeting.  

Benefits of requiring these inputs in the portal mean we have better tracking of what parameters were set for a given workflow run.  
Yes, it has the tradeoff that we may accidentally omit something, but from the data-portal side, it means we have better traceability of which samples were run with which parameters and better control of which parameters were used. 
The risk of accidentally omitting something can be mitigated but settings inputs, such as _enable_duplicate_marking_ to be mandatory, which forces us to make a conscious decision of if we set it to true or false for a workflow run.  

There are a few instances where we bake booleans into the workflow step input, but it's generally reserved for situations where it has a direct effect on a downstream step. For example, in the umccrise-with-germline-pipeline, we hardcode `enable-map-align-output` to be set to `true` otherwise the downstream umccrise step wouldn't have a bam file to work with.

alexiswl:
Let me know if you're okay with this @ohofmann and I'll put in a PR for the rest of our pipelines that have been updated in https://github.com/umccr/cwl-ica/pull/151

skanwal:
Agree with Alexis - the best way to capture inputs to workflows is via `inputs` or arguments to individual steps. Generally,  inputs to workflows is responsibility of the users. In our cases, it's tackled on the portal side, which is the main orchestrator.

ohofmann:
We discussed this in various meetings. I am not disagreeing, my point was that a) this is harder to validate and b) we have no track record of _how_ a specific workflow was run. If the CWL workflows allow parameterization at run time these parameter choices need to be documented and kept alongside the general outputs of the workflow, at least with accreditation in mind.

ohofmann:
In other words, yes, merge this PR but let's open an issue to tackle logging the workflow and selected parameters in a human readable form and storing it next to the primary workflow outputs.

alexiswl:
> we have no track record of how a specific workflow was run

I would argue that this way, we have easier access to the records since all of the input parameters for a workflow run will be tracked in the portal.  

> In other words, yes, merge this PR but let's open an issue to tackle logging the workflow and selected parameters in a human readable form and storing it next to the primary workflow outputs.

Agree that this is the best solution. I think Sehrish mentioned a while ago about the provenance tool in cwltool. This is very likely out of scope for ICAv1 but could be a feature request for ICAv2. All they'd need to do would be add the --provenance parameter to the cwltool invocation. I'm not sure how well maintained this component of cwltool is and how much information we'd be able to retrieve from this. I also don't know how well this parameter is supported on other platforms, such as AGC.

skanwal:
@alexiswl - right, the flag has its limitations. For example - all inputs get coped to  the `data/` folder of the provenance output directory. Particularly problematic, when the inputs are big or contain reference to databases. Some work is under progress and actively being discussed as well https://github.com/common-workflow-language/cwltool/issues/1586. 
Definitely worth exploring in icav2 - lets discuss further is our next 1-1.

alexiswl:
Thanks for this, looks like this won't be as simple as we would have hoped. 
Wonder if there's a way where we can help kick the tires a bit.

victorskl (commented):
Thanks folks for healthy discussion and rigorous reviews/counter checking the PR here. That's the team spirit! 💪 

Like Oliver comment; I'd recommend to track appropriate issue ticket at respective upstream repo for any further action needs; pivoting discussion here as starter context. We always have another iteration. :)


@ohofmann @reisingerf If all good, can I have 'approve' tick, please. So that I can conclude merge this PR. 

Please note label flags that I will carry out those action items accordingly.

reisingerf (approved):
Technically good!

ohofmann (approved):

victorskl:
Nice! Merging...

PR #268, opened 2022-11-08
Portal TF: Added Staging support for ICA automated pipeline

victorskl:
* Get us started Staging with replica of Production settings
  that is to be released next, as-is
* Refactored terraform code into break out by workflow types
  so that it tracks better on changes, minimise side effects
* Improved input template by decomposing into actual JSON file
  for lint, syntax and diff ready checking


alexiswl:
Does this mean a bit of duplicate handling, are there times where the production workflow ID and version will not be the same as the staging workflow ID and version? Is there a way to link the staging to the production instead?

i.e

instead of 

```
locals {
  bcl_convert_wfl_id = {
    dev  = "wfl.59e481580c6243b6b237ca2b08fa1270"
    prod = "wfl.f257ca35ced94e648fdda1173144c476"
    stg  = "wfl.f257ca35ced94e648fdda1173144c476"
  }

  bcl_convert_wfl_version = {
    dev  = "3.7.5"
    prod = "3.7.5--0bac6d0"
    stg  = "3.7.5--0bac6d0"
  }

  bcl_convert_wfl_input = {
    dev  = file("${path.module}/template/bcl_convert_dev.json")
    prod = file("${path.module}/template/bcl_convert_prod.json")
    stg  = file("${path.module}/template/bcl_convert_stg.json")
  }
}
```

Something like

```
locals {
  bcl_convert_wfl_id = {
    dev  = "wfl.59e481580c6243b6b237ca2b08fa1270"
    prod = "wfl.f257ca35ced94e648fdda1173144c476"
  }

  bcl_convert_wfl_version = {
    dev  = "3.7.5"
    prod = "3.7.5--0bac6d0"
  }

  bcl_convert_wfl_input = {
    dev  = file("${path.module}/template/bcl_convert_dev.json")
    prod = file("${path.module}/template/bcl_convert_prod.json")
    stg  = file("${path.module}/template/bcl_convert_stg.json")
  }
}

bcl_convert_wfl_id["stg"] = bcl_convert_wfl_id["prod"]
bcl_convert_wfl_version["stg"] = bcl_convert_wfl_version["prod"]
```

Understand the inputs need to be different since the templates will have references accessible in separate projects.

victorskl:
Yep; thinking similar. Happy to adapt the change proposal if that reflect the situation better, yes. 

But. Let also, open the question to the floor. See, what other team member think.

---

> Understand the inputs need to be different since the templates will have references accessible in separate projects.

... was also thinking; whether we could/should have single centralised reference data that can be pointed from differing execution environment. But. May be, that one for future, I reckon.

If we go with replica as in PR, probably I would need your help prep-ing/copying those ref data around to destination `gds://staging`...

skanwal:
> are there times where the production workflow ID and version will not be the same as the staging workflow ID and version?

Hmm, thinking about a situation where a workflow version is being tested in staging, but the results aren't satisfactory or needs enhancement. In this case, we might not want to push this version to prod? How will this be tackled?

alexiswl:
> In this case, we might not want to push/use this version in prod? How will this be tackled?

Yep good question. The merge in of a PR doesn't mean that the values are necessarily reflected in AWS. 
I think we'd first run `terraform apply` in the staging account and then if those results are satisfactory run `terraform apply` in the production account (without the need for an additional PR). 


> If we go with replica as in PR, probably I would need your help prep-ing/copying those ref data around to destination

Might be time for you to install ica-ica-lazy @victorskl !

skanwal:
> I think we'd first run terraform apply in the staging account

And from the `locals` below, if we get rid of `stg`, will terraform apply use `prod` details?

```
locals {
  bcl_convert_wfl_id = {
    dev  = "wfl.59e481580c6243b6b237ca2b08fa1270"
    prod = "wfl.f257ca35ced94e648fdda1173144c476"
    stg  = "wfl.f257ca35ced94e648fdda1173144c476"
  }

  bcl_convert_wfl_version = {
    dev  = "3.7.5"
    prod = "3.7.5--0bac6d0"
    stg  = "3.7.5--0bac6d0"
  }

  bcl_convert_wfl_input = {
    dev  = file("${path.module}/template/bcl_convert_dev.json")
    prod = file("${path.module}/template/bcl_convert_prod.json")
    stg  = file("${path.module}/template/bcl_convert_stg.json")
  }
}
```

alexiswl:
Yes if the keys point to the same value, I removed the staging lines but put in the following below that dictionary

```
bcl_convert_wfl_id["stg"] = bcl_convert_wfl_id["prod"]
bcl_convert_wfl_version["stg"] = bcl_convert_wfl_version["prod"]
```

reisingerf:
Hm... This is the cycle I had in mind (happy to discuss):

Initially `stg` will be an identical copy of `prod`. Then development will happen and be tested on `stg` (once it's reached RC status in `dev`), at which point it will start to differ from `prod`. That difference is usually reflected in different workflow ids/versions (and possibly reference data and/or code), so that should be configurable and deployable for `stg` only.
Once a certain stage of `stg` has been reached and is validated/approved, the `stg` config is copied over to `prod` to become the new production release (which will put them back in sync and the cycle repeats).

With that in mind, I'd keep the `stg`  and `prod` settings separate and independent.

Does that make sense?

alexiswl:
> Does that make sense?

So we end up with three PRs, one for dev, staging and prod?

victorskl:
> SMS parameters

😆 

Let have a brief chat in today DevOps call, also. Ok?

alexiswl (approved):

reisingerf (approved):

pdiakumis (approved):

skanwal (approved):

skanwal:
Don't know what the consensus in the call was, approved nonetheless.

victorskl:
Yep. No worries, Sehrish. We agree to go ahead explicit config as-in PR for now. And we shall revisit/review again in next few iterations down the track. 

As this is still on the basis of, such that, we are still endeavouring into this new release approach. So, not sure what might work better yet nor moving parts. So being explicit config setting to begin with our journey!

Join us sometime in DevOps. We miss ya! 😉

victorskl (commented):
Thanks team for healthy discussion! 🙂 Going to merge this now. And, me to proceed next step.

PR #269, opened 2022-11-10
Portal TF: Refactored Portal baseline resources

victorskl:
* There is no changes in functional wise. Just technical rearrangement
  and break out of `main.tf` module into respective cohesive resources.


reisingerf (approved):
Go for it!

PR #271, opened 2022-11-18
Initial commit of umccr_data stack

reisingerf:
Start to setup our data account.
For now, only meant for S3 buckets to hold archive data.

### DIFF
--- a/terraform/stacks/data_archive/main.tf
+++ b/terraform/stacks/data_archive/main.tf
@@ -1,0 +1,110 @@
+terraform {
+  required_version = "~> 1.3.3"
+
+  backend "s3" {
+    # AWS access credentials are retrieved from env variables
+    bucket         = "umccr-terraform-states"
+    key            = "data_archive/terraform.tfstate"
+    region         = "ap-southeast-2"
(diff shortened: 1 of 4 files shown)


terraform/stacks/data_archive/main.tf
```diff
@@ -0,0 +1,110 @@
...
+      sse_algorithm = "AES256"
+    }
+  }
+}
+
+resource "aws_s3_bucket_versioning" "raw_data_bucket" {
+  bucket = aws_s3_bucket.raw_data_bucket.id
+
+  versioning_configuration {
+    status = "Enabled"  ## Should we disable versioning?
```
victorskl, line 77: > Should we disable versioning?

I think so. Similar to raw data bucket policy?

victorskl (approved):
LGTM!

terraform/stacks/data_archive/main.tf
reisingerf: Yes. For now only testing.
I'll leave it on, as it should not hurt, but could safe some trouble in the beginning. 
Can change/revise when this is going to get used in a production setting.

Issue #272, opened 2022-11-21
Migrate to new Portal domain name


victorskl:
**Context:**

- Now that we have dedicated "data" account; and to avoid confusion and better representation purpose, it is proposing change from `data.umccr.org` to `portal.umccr.org`

**Actions:**

- To discuss in team meetings
- We shall do this in stages across period of time (i.e. taking time for graceful transition; aiming around Q2/Q3 '23)

victorskl:
Update 09/12/2022

https://umccr.slack.com/archives/C025TLC7D/p1670580816232859

victorskl:
Status Page UI updated Portal domain usage with PR https://github.com/umccr/data-portal-status-page/pull/83

victorskl:
We now totally removed DNS entries (first pass - soft deprecation) from Route53.

Announced in Slack - https://umccr.slack.com/archives/C025TLC7D/p1721208827283479

In next Portal update, I will totally remove (old) Portal UI CodePipeline deployment from terraform / serverless.

Schedule with [2023.02-b3](https://github.com/umccr/infrastructure/milestone/13)

victorskl:
Superseded by Portal to OrcaBus system migration.

See

https://umccr.slack.com/archives/CP356DDCH/p1728255673587009
https://umccr.slack.com/archives/C025TLC7D/p1743567893300159
https://umccr.slack.com/archives/C025SCF86/p1743036301961559

victorskl:
Consider release portal domain so that it can be used (DNS re-pointing) for OrcaUI Cloudfront.

https://umccr.slack.com/archives/C025TLC7D/p1729052619637879?thread_ts=1729051643.526059&cid=C025TLC7D

victorskl:
Mark 2 years epic ticket sprint. Legacy Gen-1 Portal is down, finally!

PR #273, opened 2022-11-22
Portal & Cognito AAI TF: Added dual frontend for serving Portal v1 & v2

victorskl:
* Internally codename as `data2` client that would use the same Portal backend
  but with newer Portal v2 UI.
* Client repo branch `v1-main` is still current PROD at data.umccr.org
* Client repo branch `main` merges will pipe to portal.umccr.org
* Also started of #272 and, together with v2 UI migration change management time
  window towards Q2/Q3 '23 eta


reisingerf (approved):
Nice!

williamputraintan (approved):
LGTM

PR #279, opened 2023-02-13
New app: Dracarys

brainstorm:
Related with PR https://github.com/umccr/data-portal-apis/pull/500 since SSM resources (SQS queue and S3 bucket) need to be instantiated.


terraform/stacks/umccr_data_portal/pipeline/main.tf
```diff
@@ -224,4 +224,25 @@ resource "aws_ssm_parameter" "sqs_somalier_extract_queue_arn" {
...
   value = aws_sqs_queue.somalier_extract_queue.arn
   tags  = merge(local.default_tags)
-}
\ No newline at end of file
+}
+
+# TODO: There's no clear guidance/documentation about where to plug new "apps"
+# in the (data-portal pipeline?): it's either in here or in app/ dir?
+#
+# This needs to be documented!
```
victorskl, line 232: Yup. Need to note better. Since Dracarys integration follows WES event orchestration pattern, please use pipeline here, pls.

brainstorm:
Ready to review and first PoC up and running, I reckon:

```
% ./test_queue_msg.sh
% aws s3 ls s3://umccr-datalake-dev/portal/creation_date=2023-03-17/
2023-03-17 16:44:36       3547 dracarys_multiqc.tsv.gz
```

Things that need immediate attention before merging:

1. Handling of other Dracarys tidied inputs.
2. Narrow down the perms on the destination S3 bucket.
3. Integration with https://github.com/umccr/data-portal-apis/pull/551

/cc @alexiswl Would you like to try and consume that resulting file?

cdk/apps/dracarys/README.md
```diff
@@ -0,0 +1 @@
+This is a CDK repo deploying a lambda that takes in a presign URL and a prefix, runs Dracarys on the file downloaded from the presign URL, and writes the output to S3.
```
reisingerf, line 1: I thought it takes a directory as input and figures out by itself which files to look at/ingest?
Perhaps document the input parameters? 
(I know I can pull it from the code, but I am lazy ;-) )

reisingerf: And perhaps a link to `Dracarys` for context?

cdk/apps/dracarys/dracarys/dracarys_stack.py
```diff
@@ -0,0 +1,61 @@
...
+        lambda_role = aws_iam.Role(scope=self, id='dracarys-lambda-role',
+                                assumed_by =aws_iam.ServicePrincipal('lambda.amazonaws.com'),
+                                role_name='dracarys-lambda-role',
+                                managed_policies=[
+                                aws_iam.ManagedPolicy.from_aws_managed_policy_name(
+                                    'service-role/AWSLambdaBasicExecutionRole'),
+                                aws_iam.ManagedPolicy.from_aws_managed_policy_name(
+                                    'service-role/AWSLambdaVPCAccessExecutionRole'),
+                                aws_iam.ManagedPolicy.from_aws_managed_policy_name(
+                                    'SecretsManagerReadWrite'),
```
reisingerf, line 28: Why does it need write permissions?

cdk/apps/dracarys/dracarys/dracarys_stack.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Fetch dracarys queue ARN from pre-existing SSM
+        # created by terraform's data portal pipeline stack
+        queue_name = "/data_portal/backend/sqs_dracarys_queue"
+        queue_arn = aws_ssm.StringParameter.from_string_parameter_attributes(self, "dracarys_sqs_queue",
+                parameter_name=queue_name).string_value
+
+        queue = aws_sqs.Queue.from_queue_arn(self, id="data-portal-dracarys-queue", queue_arn=queue_arn)
+
+        bucket = aws_s3.Bucket(self, "umccr-datalake-dev")
```
reisingerf, line 41: I suppose the bucket name would be parameterised in the future?

reisingerf (approved):
Some minor comments, but overall LGTM

cdk/apps/dracarys/dracarys/dracarys_stack.py
brainstorm: Couldn't find an AWS managed for SecretsManagerRead(Only) quick and didn't want to diy a non-managed policy, but definitely needs fixing :)

cdk/apps/dracarys/dracarys/dracarys_stack.py
brainstorm: What's the current best practice? Lambda env? SSM? I've seen this: https://github.com/umccr/infrastructure/blob/master/cdk/apps/thebus/stacks/thebus_stack.py#L35 and made me wonder 🤔

alexiswl:
> /cc @alexiswl Would you like to try and consume that resulting file?

Yes will have a go today

cdk/apps/dracarys/dracarys/dracarys_stack.py
reisingerf: I think according to the best practices [guide](https://blog.serverlessadvocate.com/serverless-aws-cdk-pipeline-best-practices-patterns-part-1-ab80962f109d) it should be neither, but a stack specific configs.

brainstorm: Went through that 2-part guide this morning (such a shame part 3 is not there yet since it deals with variables/secrets :__/). I think that following that CDK Pipelines requires requires quite some time to refactor/rework from here, what "best" practice would you think it'd be acceptable to adopt **today**?

alexiswl: > just passing props from some parent stack?

I found the parsing of properties between stacks much easier with the TypeScript implementation of CDK (if that's of any consolation). For me it was enough to switch (which I did for the cttso-ica-to-pieriandx repo).  

I've also just got a bash script which I use to sync ssm parameters with (https://github.com/umccr/cttso-ica-to-pieriandx/blob/main/deploy/cttso-ica-to-pieriandx-cdk/scripts/update-params.sh) but this is probably better handled with terraform

alexiswl: > What's the current best practice across our stacks? Lambda env? SSM? I've seen this: [master/cdk/apps/thebus/stacks/thebus_stack.py#L35](https://github.com/umccr/infrastructure/blob/master/cdk/apps/thebus/stacks/thebus_stack.py?rgh-link-date=2023-03-20T00%3A19%3A30Z#L35) and made me wonder 🤔

I have a constants file I use to import variables like this

https://github.com/umccr/cttso-ica-to-pieriandx/blob/main/deploy/cttso-ica-to-pieriandx-cdk/constants.ts 

Similar to holmes (@andrewpatto)

pdiakumis:
@alexiswl there are also a ton of tidied MultiQC _Parquet_ files under prod in `s3://umccr-datalake-prod/v1/multiqc/dragen_umccrise/year=2022/month=09/` (go further up for additional workflows). Feel free to copy some of those out somewhere if prod doesn't suit.

reisingerf:
IMHO the output of Dracarys could be tsv.
It's easier to read manually and there are usually only one or two columns (one for each lib, right?). Not sure parquet would be a good fit here. (that's the data lake raw data level though)

On the data lake side I suppose a crawler process would then aggregate/clean those into (delta) tables? Backed by parquet? 
(as that's the data lake side, and should be re-producible from the raw data (tsv) files, it's up to the data lake which format to store it's tables in and it's not fixed/can change)

Does that make sense?

brainstorm:
> IMHO the output of Dracarys could be tsv. It's easier to read manually and there are usually only one or two columns (one for each lib, right?). Not sure parquet would be a good fit here. (that's the data lake raw data level though)
> 
> On the data lake side I suppose a crawler process would then aggregate/clean those into (delta) tables? Backed by parquet? (as that's the data lake side, and should be re-producible from the raw data (tsv) files, it's up to the data lake which format to store it's tables in and it's not fixed/can change)
> 
> Does that make sense?

Sorry to put my foot here, but nowadays it makes sense to transition to parquet for several reasons:

1) Storage and speed savings.
2) It is supported by default as an ingestion format for Athena, Pandas, R and pretty much all ecosystems.

See one of the many articles that advocate for this transition: https://datapythonista.me/blog/pandas-20-and-the-arrow-revolution-part-i

Specially in bioinformatics, the metadata can get big and there are fewer and fewer reasons (today) to still stick with TSV/CSV formats.

cdk/apps/dracarys/dracarys/dracarys_stack.py
brainstorm: Cool, thanks Alexis, I think I might just go for the constants route for now, thanks for chiming in!

reisingerf:
Hm, I get your point and I agree in principle. However, I think we are talking about different use cases here, or I misunderstood it. If I did, please ignore!

I think we are talking about Dracarys output, which if I recall correctly (Peter please correct me!) consists of one or two rows of metrics (one row per sample/library)? If we would talk about 100s or 1000s of rows, then yes absolutely.

Also, the idea is that we don't operate directly on those files, but they are ingested and data aggregated (which then should definitely not be in tsv).

And why parquet and not avro, orc, arrow, or something else that's possibly better suited for future use cases?
I am not super familiar with parquet et al, but I looked at (tried to anyway) some of those files with a default text editor... not always a nice experience... And no, I don't want to break out Python or R to just see if a single value is present in a single file.
(also thinking about default linux tooling, like `less`, `head`, `vim`, etc when running a quick debugging job from an EC2 instance, or the simple text display possible via the portal)
(and I completely discard Excel compatibility ;-) )

Your point 3 about included types would be useful, but I am not sure it's that critical?

alexiswl:
@reisingerf I agree, the efficiencies for parquet surely come into play with 100s or 1000s of rows but we have datasets here that are one row with a few hundred columns. I certainly had no issues ingesting either but would probably have a lot less trouble parsing a compressed TSV through the command line compared to a parquet file.

brainstorm:
Alright, I'll create both formats on different S3 prefixes, [see here](https://github.com/umccr/infrastructure/commit/e9bb7a89097fb3b270d3b3582179b74b9645124f#diff-8f4808e7159a7778740b25ff7eb07fcb804114281e9721024c70efbbf6edfe3fR47-R56) ;)

PR #281, opened 2023-02-15
Htsget CDK: Final tear down note on htsget-refserver deprecation

### LINKED ISSUE #214: Migrate to htsget-rs
**Context:**

- Our existing [htsget deployment](https://github.com/umccr/infrastructure/tree/master/cdk/apps/htsget) use reference implementation (RI instances)

**Actions:**

- Migrate it to [htsget-rs](https://github.com/umccr/htsget-rs)
- Then, tear down (now obsolete) RI instances from both DEV and PROD

victorskl:
Resolves #214 #278


brainstorm:
@victorskl I'd probably remove this app altogether since it'll be present in the github history anyway? Reduce clutter?

victorskl:
Yeah. That's also do it, actually. So, I will update the PR with rm -rf, ok?

victorskl:
Also. We should remove this entry too -- https://github.com/umccr/wiki/tree/master/bioinformatics/htsget

The procedure is a bit different now with htsget-rs.

I think it is better we remove them; in order to avoid confusion for end user.

You are right, we can always go back dig git history, if any.

If you are agree, I will carry the cleanup, tick next. @brainstorm

brainstorm:
Next! :P

victorskl:
Gone now, Roman. Need your tick approval for both PRs!

PR #283, opened 2023-03-03
GPL CI/CD URL Removal

williamputraintan:
GPL FuncUrl no longer needed at the portal and will call lambda invoke instead. (https://github.com/umccr/data-portal-client/issues/235)

Ref: https://github.com/umccr/data-portal-client/pull/237#pullrequestreview-1322967240

### DIFF
--- a/terraform/stacks/umccr_data_portal/.terraform.lock.hcl
+++ b/terraform/stacks/umccr_data_portal/.terraform.lock.hcl
@@ -5,6 +5,7 @@ provider "registry.terraform.io/hashicorp/aws" {
   version     = "4.38.0"
   constraints = "4.38.0"
   hashes = [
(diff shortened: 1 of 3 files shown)


victorskl:
> Is there any useCase of functionUrl aside from portal? else consider removing it from the GPL stack?

Good observation; Will. We shall keep functionUrl at GPL side for [Snowshoe](https://github.com/umccr/infrastructure/milestone/9) release cycle. Just in case we need to revert during this time.

And. If all things go well; probably we shall phase out at GPL side in next tick, [Ocicat](https://github.com/umccr/infrastructure/milestone/10), if any.

@scwatts FYI

victorskl (approved):
LGTM, thanks!

victorskl:
@williamputraintan Pls merge, if no more commits; thkz

PR #284, opened 2023-03-06
Update dev workflow ids for somatic and umccrise

alexiswl:
Dragen somatic now also calls germline tool

Runs germline in somatic step, not umccrise step

PR #286, opened 2023-03-12
Update umccrise dev workflow version

alexiswl:
Forgot to change the workflow version in #284 

This fixes the workflow version for umccrise-workflow only versions


victorskl (approved):
Thanks Alexis. I will merge it and apply to dev, next.

PR #305, opened 2023-05-05
Updated staging workflow versions and template for tn and umccrise

alexiswl:
Upgraded versions of tn pipeline and umccrise tool to allow for bam to not be generated in germline directory.  

This requires a umccrise version upgrade to copy over the normal bam to the umccrise directory first.

### DIFF
--- a/terraform/stacks/umccr_data_portal/workflow/template/wgs_tumor_normal_stg.json
+++ b/terraform/stacks/umccr_data_portal/workflow/template/wgs_tumor_normal_stg.json
@@ -6,8 +6,10 @@
   "fastq_list_rows": [],
   "tumor_fastq_list_rows": [],
   "enable_map_align_output": true,
+  "enable_map_align_output_germline": false,
   "enable_duplicate_marking": true,
   "enable_sv": true,
+  "enable_sv_germline": false,
   "reference_tar": {
     "class": "File",
     "location": "gds://staging/reference-data/dragen_hash_tables/v8/hg38/altaware-cnv-anchored/hg38-v8-altaware-cnv-anchored.tar.gz"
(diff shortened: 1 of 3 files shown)


victorskl:
Thanks @alexiswl -- Similar to house keeping comment [here](https://github.com/umccr/data-portal-apis/pull/588#issuecomment-1537647843) 

Could you pls link/create related Trello card for this, pls?

Me, need to link from Trello to this PR; to coordinate changes to production.

I will review the PR soon in tick.

terraform/stacks/umccr_data_portal/workflow/template/wgs_tumor_normal_stg.json
```diff
@@ -6,8 +6,10 @@
   "fastq_list_rows": [],
   "tumor_fastq_list_rows": [],
   "enable_map_align_output": true,
+  "enable_map_align_output_germline": false,
   "enable_duplicate_marking": true,
   "enable_sv": true,
+  "enable_sv_germline": false,
```
victorskl, line 12: So, just to ask/confirm; this two flags signal not to produce that extra BAM, correct?

victorskl (approved):
LGTM! Just a question only, otherwise. Happy to apply it to STG, once merged.

terraform/stacks/umccr_data_portal/workflow/template/wgs_tumor_normal_stg.json
alexiswl: We set enable_map_align_output to false for dragen germline to not produce the extra bam.

enable_sv_germline needs to be set to false because enable_map_align_output needs to be set to be true in order for enable_sv to be set to be true.

victorskl: Thanks for explain. I will merge this, tick next. And, as for next action, please kindly help check in Trello card.

PR #319, opened 2023-06-06
Data: Add ICAv2 BYOB bucket resources

reisingerf:
This will add a new bucket and IAM user to be used with ICAv2.
Questions welcome!

### DIFF
--- a/terraform/stacks/data_archive/byob_ica_v2.tf
+++ b/terraform/stacks/data_archive/byob_ica_v2.tf
@@ -1,0 +1,194 @@
+################################################################################
+# Variables
+
+variable "analysis_data_bucket" {
+  description = "This bucket is holding production data."
+  default     = "org.umccr.data.analysis_data"
+}
+variable "analysis_data_prefix" {
+  description = "The prefix under which all analysis data is collected."
+  default     = "analysis_data/"
+}
+
+################################################################################
+# Buckets
+
+resource "aws_s3_bucket" "analysis_data_bucket" {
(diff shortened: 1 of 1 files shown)


victorskl:
[ICA V2 Migration: Test multiple domain approach](https://trello.com/c/jqPvbzQs/1336-ica-v2-migration-test-multiple-domain-approach)

victorskl (approved):
LGTM!

alexiswl:
I'm not a terraform expert, looks fine, I think? 

Surprised that's the only permissions you need for the bucket? If V2 objects are immutable, is that a layer that's only protected in V2?

alexiswl (approved):

reisingerf:
>Surprised that's the only permissions you need for the bucket?

I am just following their docs and so far it seems to work.
https://help.ica.illumina.com/home/h-storage/s-awss3
(just noticed I missed the CORS bit, which is only needed for browser based operations)

>If V2 objects are immutable, is that a layer that's only protected in the V2 API? But if one was to go via S3? Could they make a new object version?

Well... yes and no, I think. 
Yes, in the sense that they can only control what's on their side. It's our bucket and we retain ownership and full control, so we can bypass whatever controls they put in place. 
No, in the sense that control doesn't only have to be on their side. If bucket versioning is enabled, the objects don't get deleted (unless expired or otherwise cleaned up). And the role they use does not have permissions over object versions, so they can't "force delete" an object.

victorskl (approved):
LGTM too! CORS config looks fine.

PR #323, opened 2023-06-21
Data/byob update

reisingerf:
FYI @victorskl 
This is to reflect that one ICA project can only have one BYOB bucket configured, but we could use the same bucket across multiple projects. So the prefixes should reflect that (and also provision for data not under ICA control).

### DIFF
--- a/terraform/stacks/data_archive/byob_ica_v2.tf
+++ b/terraform/stacks/data_archive/byob_ica_v2.tf
@@ -1,37 +1,29 @@
 ################################################################################
-# Variables
+# Local constants
 
-variable "production_data_bucket" {
-  description = "This bucket is holding production data."
-  default     = "org.umccr.data.production"
-}
-variable "ica_byob_prefix" {
-  description = "The prefix linked to ICAv2 BYOB configuration."
-  default     = "icav2/byob/"
-}
-variable "production_data_prefix" {
-  description = "The prefix linked to ICAv2 'production' project'."
-  default     = "${var.ica_byob_prefix}production/"
-}
-variable "production_temp_data_prefix" {
-  description = "The prefix for temporary data in the ICAv2 'production' project. Subject to lifecycle management."
-  default     = "${var.production_data_prefix}temp/"
-}
-variable "production_analysis_data_prefix" {
-  description = "The prefix for analysis data in the ICAv2 'production' project Subject to lifecycle management."
-  default     = "${var.production_data_prefix}analysis-data/"
+locals {
+  # The bucket holding all "production" data
+  production_data_bucket = "org.umccr.data.production"
+  # prefix used for the ICAv2 BYOB config (mapped to and accessible from ICA)
+  ica_byob_prefix = "icav2/byob/"
+  # prefix for data in the ICA "production" project
+  production_data_prefix = "${local.ica_byob_prefix}production/"
+  # prefix for analysis data in the ICA "production" project, subject to lifecycle management
+  production_temp_data_prefix = "${local.production_data_prefix}temp/"
+  # prefix for temporary data in the ICA "production" project, subject to lifecycle management
+  production_analysis_data_prefix = "${local.production_data_prefix}analysis-data/"
 }
 
 ################################################################################
 # Buckets
 
 resource "aws_s3_bucket" "production_data_bucket" {
-  bucket = var.production_data_bucket
+  bucket = local.production_data_bucket
 
   tags = merge(
     local.default_tags,
     {
-      "Name"=var.production_data_bucket
+      "Name"=local.production_data_bucket
     }
   )
 }
@@ -88,7 +80,7 @@ resource "aws_s3_bucket_lifecycle_configuration" "production_data_bucket" {
     id = "analysis_data_rule"
     status = "Enabled"
     filter {
-      prefix = var.production_analysis_data_prefix
+      prefix = local.production_analysis_data_prefix
     }
     transition {
       days          = 0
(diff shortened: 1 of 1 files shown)


terraform/stacks/data_archive/byob_ica_v2.tf
```diff
@@ -1,41 +1,45 @@
...
+
+locals {
+  # The bucket holding all "production" data
+  production_data_bucket = "org.umccr.data.production"
+  # prefix used for the ICAv2 BYOB config (mapped to and accessible from ICA)
+  ica_byob_prefix = "icav2/byob/"
+  # prefix for data in the ICA "production" project
+  production_data_prefix = "${local.ica_byob_prefix}production/"
+  # prefix for analysis data in the ICA "production" project, subject to lifecycle management
+  production_temp_data_prefix = "${local.production_data_prefix}temp/"
```
alexiswl, line 12: I'm not sure we'll need a temp folder in the same way we need one for v1. Unless this is v1 we are referring to here?

reisingerf: There won't be temporary workflow output? 

Note: this new configuration means ALL production data would be in this bucket, also from pipelines run outside of ICA. It's basically an active "work" space for all data that's "currently" needed. All data is then supposed to be either deleted or moved to an archive.

reisingerf: Did you have a play with the cross project volume/data mapping options? 
It seems possible, but not sure it's good enough for production work, i.e. I am not sure I want to rely on it as key part of the production setup.

alexiswl: > Did you have a play with the cross project volume/data mapping options?

Not sure what you mean by this, 
As in just creating files in one project bundle and sharing it to another project?

alexiswl: > There won't be temporary workflow output?

Nope, either we get the outputs (for a successful workflow) or we don't. All temp data is on the local filesystem and then the out/ directory is uploaded after completion of the workflow back to S3

reisingerf: >As in just creating files in one project bundle and sharing it to another project?

Well, the alternative to using a single bucket (which is linked to the `production` project) is to make other buckets/volumes available to the `production` project. 
Say a workflow reads oncoanalyser data, runs some analysis/reporting and writes its result back out. If all workflows (ICA and other) use the same bucket, there is no issue with linking/access. If oncoanalyser uses a different bucket, the data would not automatically be available in the `production` project's linked storage (various options for making it available... ). Things get even more complex if we wanted to keep "research" separate from "clinical" data. 

Probably manageable, but we'd probably end up with a combination of settings to make data available (pre-signed URLs, project/data linking, shared/multiple project access, ...) and I wonder if that's worth it. I have not clear feeling for all the issues/limitations we'd likely run into.

The simpler alternative is to have ALL "active" data in the same bucket and manage it there. A bit more risk as everyone/thing will need to have access, but at least that's mainly on our end/in our control. 

So the question here is: one bucket to rule them all or divide and conquer?

reisingerf: >Nope, either we get the outputs (for a successful workflow) or we don't.

I am happy to remove it if there are no use cases, but we have this in the current ICA setup/volume. 
I guess it does not hurt to have a location ready just in case?

alexiswl: > I guess it does not hurt to have a location ready just in case?

Sure no harm in keeping it.  

> The simpler alternative is to have ALL "active" data in the same bucket and manage it there

I think this is the right approach. We can have validation and reference data in a different bucket linked to a different ICAv2 project.  

> A bit more risk as everyone/thing will need to have access, but at least that's mainly on our end/in our control.

It is a bit more risk. But we could lock things down almost entirely on the ICA side -- and use it almost exclusively as a workflow execution service rather than a data access point.  

Also reducing time between data creation and archive date would reduce the risk in the worst case scenario that the entire bucket is compromised. In ICAv1, we essentially have all data in the one key prefix (and it's all active), so our own single BYOB is still better than the current setup.

reisingerf: Agreed!

alexiswl (approved):

terraform/stacks/data_archive/byob_ica_v2.tf
```diff
@@ -1,41 +1,45 @@
...
-variable "analysis_data_prefix" {
-  description = "The prefix under which all analysis data is collected."
-  default     = "analysis_data/"
+# Local constants
+
+locals {
+  # The bucket holding all "production" data
+  production_data_bucket = "org.umccr.data.production"
+  # prefix used for the ICAv2 BYOB config (mapped to and accessible from ICA)
+  ica_byob_prefix = "icav2/byob/"
```
reisingerf, line 8: Since we agreed to go with the one bucket approach, I guess this prefix does not make a lot of sense any more.
We could keep the storage separated between execution engines, but that goes a little against the "all active production data in one place" philosophy (and may again create access/sharing complications).

We could simply expose the whole bucket. Not sure if we'd be limiting our options... 
Thoughts? 
@alexiswl @victorskl

victorskl: A quick call, pls?

alexiswl: > We could keep the storage separated between execution engines

Happy for them to share, I think it would be a nice-to-have if we could make sure all workflow execution points can output data in a similar format (so it'd essentially be impossible to tell which WES service created a directory).

victorskl:
[ICA V2 Migration: Test multiple domain approach](https://trello.com/c/jqPvbzQs/1336-ica-v2-migration-test-multiple-domain-approach)

victorskl (approved):
LGTM!

alexiswl (approved):

PR #349, opened 2023-08-23
Feature/cdk bootstrap via terraform

andrewpatto:


victorskl:
[UoM AWS Migration: Bootstrap CDK in new UoM AWS accounts](https://trello.com/c/jn56wL6f/1439-uom-aws-migration-bootstrap-cdk-in-new-uom-aws-accounts)

terraform/stacks/cdk_bootstrap/main.tf
```diff
@@ -0,0 +1,39 @@
...
+}
+
+resource "aws_cloudformation_stack" "CDKToolkit" {
+  name = "CDKToolkit"
+  capabilities = [
+    "CAPABILITY_NAMED_IAM"
+  ]
+
+  parameters = {
+    CloudFormationExecutionPolicies = "arn:aws:iam::aws:policy/AdministratorAccess"
```
victorskl, line 30: Perhaps, we shall also start creating a dedicated IAM role in this TF stack for this role policy..? And, start tracking / updating here as need be basis - expanding CF exec permission policy as we get go..? 

I like the idea proposed in this article, as a starter.
https://betterdev.blog/cdk-bootstrap-least-deployment-privilege/

victorskl (approved):
I like this.! LGTM for first pass. Minor comments for discussion/further enhancement, if agree.

reisingerf:
Happy with this, but I would also have been happy with a simple shell script wrapping the bootstrap.
This now means that TF has to be bootstrapped before you can bootstrap CDK, right? 
(and I don't think we have scripted TF bootstrapping... yet)

terraform/stacks/cdk_bootstrap/main.tf
andrewpatto: Yes - I think there is probably a best practice approach to the IAM role and the permissions boundaries - that I guess we can work towards.
AWS seems to be doing some approach to the use of permission boundaries in the CDK bootstrap (there are extra parameters now).. but seems like a work in progress

terraform/stacks/cdk_bootstrap/main.tf
```diff
@@ -0,0 +1,39 @@
...
+    FileAssetsBucketKmsKeyId        = "AWS_MANAGED_KEY"
+    PublicAccessBlockConfiguration  = true
+    // trust the umccr-build account for CDK pipeline deployments
+    TrustedAccounts                 = "843407916570"
```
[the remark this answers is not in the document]
andrewpatto: yes, comma separated list of accounts - but just this if only one.

Technically, no need to have a trusted account when boostrapping the toolchain account itself - but I don't think it matters if its lists itself

victorskl:
@reisingerf @andrewpatto Flo/Patto; shall we at least merge this for now..? 

Agree that this is still moving target and, not set-in-stone. Happy to follow up with enhancement approach in future iteration. Okay..?

Flo; tick pls.

reisingerf (approved):

PR #307, opened 2023-05-11
Allow cross-account-sharing for star-align-nf ECR

alexiswl:

PR #299, opened 2023-04-18
Update bastion ecr permissions for oncoanalyser repo

alexiswl:
Copied policies used for the cttso-ica-to-pieriandx tf stack above

Allows images in the bastion ecr repository 'oncoanalyser' to be accessed from staging and production

PR #357, opened 2023-08-30
Add sash ECR repo in Bastion and allow across account access

scwatts:
Change overview:

* adds an ECR repo for sash in the Bastion account
* attaches existing policy for across account sharing/access

Also see:

* https://github.com/umccr/infrastructure/pull/299/
* https://github.com/umccr/infrastructure/pull/307

### DIFF
--- a/terraform/stacks/bastion_ecr/main.tf
+++ b/terraform/stacks/bastion_ecr/main.tf
@@ -102,6 +102,24 @@ resource "aws_ecr_repository_policy" "star_align_nf_cross_accounts" {
   policy     = templatefile("policies/cross_accounts_policy_umccr.json", {})
 }
 
+
+####
+# sash
+#
+
+resource "aws_ecr_repository" "sash" {
+  name                 = "sash"
+  image_tag_mutability = "MUTABLE"
+  image_scanning_configuration {
+    scan_on_push = false
+  }
+}
+
+resource "aws_ecr_repository_policy" "sash_cross_accounts" {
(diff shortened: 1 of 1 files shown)


terraform/stacks/bastion_ecr/main.tf
```diff
@@ -102,6 +102,24 @@ resource "aws_ecr_repository_policy" "star_align_nf_cross_accounts" {
...
+  image_scanning_configuration {
+    scan_on_push = false
+  }
+}
+
+resource "aws_ecr_repository_policy" "sash_cross_accounts" {
+  repository = aws_ecr_repository.sash.name
+  policy     = templatefile("policies/cross_accounts_policy_umccr.json", {})
+}
+
```
victorskl, line 122: ah, looks like `star_align_nf_lifecycle` policy block got pushed down (i.e. as sash blocks get inserted)

Could you double check that please..?

Also, you would want `sash_lifecycle` policy for untagged images like others..?

victorskl (commented):
Minor comment to attend, pls

terraform/stacks/bastion_ecr/main.tf
scwatts: Thanks for catching that, fixed!

victorskl (approved):
Cool.! LGTM. I will merge & apply to Bastion.

victorskl:
Applied

scwatts:
Great thank you! I'll be putting it to use in an hour or so 😁

PR #370, opened 2023-10-12
Added WIP script to update accounts for bootstrapping

andrewpatto:


andrewpatto:
Another thing we could do is have some sort of "master" data structure (JSON?) naming a labelling every account.

And from the data structure we could do things like generate the appropriate aws config dynamically (would be a bit of jq).

So at the moment I think here we would be committing to keeping this master config ini file (and william has one very similar for his scripts)

andrewpatto:
But anyhow this is just a starting point

scripts/all-accounts-aws-config.ini
```diff
@@ -0,0 +1,33 @@
...
+sso_session = umccr-bootstrap-session
+sso_account_id = 843407916570
+sso_role_name = AdministratorAccess
+region = ap-southeast-2
+
+[profile unimelb-australiangenomics]
+sso_session = unimelb-bootstrap-session
+sso_account_id = 258177526432
+sso_role_name = AWSAdministratorAccess
+region = ap-southeast-2
```
andrewpatto, line 33: So if we are going to do this kind of script (also see the security scripts) - we probably should have some common master definition of all of this

victorskl: Yep. Agree, thought of the same.

victorskl (approved):
Thanks for starter, Patto. Looks good from my side. 

I hope, you could kick start in new AG account there, first, pls. We will catchup in tick after Ocicat.

andrewpatto:
Yes AG has been bootstrapped

victorskl:
[UoM AWS Migration: Bootstrap CDK in new UoM AWS accounts](https://trello.com/c/jn56wL6f/1439-uom-aws-migration-bootstrap-cdk-in-new-uom-aws-accounts)

scripts/all-accounts-aws-config.ini
andrewpatto: A master JSON that we can JQ into various forms??

victorskl: Sure

reisingerf:
Awesome, thanks!
I agree, there's lots of potential to build on. I think sharing the same config is a good first step that also promotes adoption of the same naming convention.
I think we may also be able to list all accounts we have access to using the SSO, so could dynamically discover and bootstrap them...
Account management and initial setup is also a topic we have brought up with UoM, but never resolved. Perhaps they'd be happy to do the bootstrapping for us (I'm probably dreaming here).

reisingerf (approved):

PR #375, opened 2023-10-17
Created three stacks for development for testing the icav2 credentials

alexiswl:
(for the umccr-test user account)

Waiting on https://github.com/umccr/cwl-ica/pull/407

Also need to credit with previous links to ica_credentials folder as most of the code has come from there.  

Changes include:
* 1-1 mapping between ApiKey and token 
  * since v2 does NOT let us scope tokens or keys, each stack represents a different service account
* Addition of github role and repos.  For the pipeline stack, we create a github role which allows cwl-ica access to create pipelines.
  * Note user will need access to reference / validation data.

### DIFF
--- a/cdk/apps/icav2_credentials/.gitignore
+++ b/cdk/apps/icav2_credentials/.gitignore
@@ -1,0 +1,8 @@
+*.js
+!jest.config.js
+*.d.ts
+node_modules
+
+# CDK asset staging directory
+.cdk.staging
+cdk.out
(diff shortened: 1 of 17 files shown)


alexiswl:
Checked - not a secret 
image

victorskl:
Might want to add the following pointer in README to GitHub doc about setup

https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services

victorskl (approved):

PR #386, opened 2023-10-30
Slack CDK: Improved multi-accounts slacker

victorskl:
* Upgraded to Python3.11 runtime
* Upgraded to CDK v2
* Refactored CDK Python project setup
  (not a lib project so requirements.txt is just fine)
* Locked, pinned, localised project scoped CDK CLI version
  (global cdk cli can be drifted away from project scope)
* Marked deprecated stacks for Stage1 phasing out
  (so that we can remove them for good in next update)
* Improved README - how to dev/ops/use the stack


victorskl:
Tag as part of Ocicat as it has used in Oncoanalyser Slack notification purpose. Still pending apply prod.

reisingerf (approved):

PR #388, opened 2023-12-04
AG's account TF update

williamputraintan:
- Removal of `s3://agha-gdr-mm` as it is no longer for use /  data inside is copied to gdr-store-2 bucket
- Re-import lifecycle configuration policy for MM deep archive
- Reinforce bucket owner object policy on all 3 active AG's buckets

Todo next:
- removal in IAM user credential for AGHA user

* TF is not applied yet


terraform/stacks/agha_infra/main.tf
```diff
@@ -132,6 +139,67 @@ resource "aws_s3_bucket" "agha_gdr_store_2" {
...
+
+  rule {
+    status = "Enabled"
+    id = "[DELETE NonCurrVer] MM_NSWHP/2022"
+    
+    filter {
+      prefix = "MM_NSWHP/2022"
+    }
+
+    noncurrent_version_expiration {
```
reisingerf, line 183: Merge rules that apply to the same prefix? 
Or set this globally on the whole bucket (would we ever have prefix specific expiration settings?)

terraform/stacks/agha_infra/main.tf
```diff
@@ -83,11 +85,16 @@ resource "aws_s3_bucket_public_access_block" "agha_gdr_staging_2" {
   ignore_public_acls      = true
   restrict_public_buckets = true
 }
+resource "aws_s3_bucket_ownership_controls" "agha_gdr_staging_2" {
```
reisingerf, line 88: Why do we need this? 
(I think it's against the best practises)

terraform/stacks/agha_infra/main.tf
```diff
@@ -75,6 +76,7 @@ resource "aws_s3_bucket" "agha_gdr_staging_2" {
     }
   )
 }
+
 resource "aws_s3_bucket_public_access_block" "agha_gdr_staging_2" {
```
reisingerf, line 80: Do we still need an explicit public access block setting?
(this should be the default now, no?)

reisingerf (commented):
Overall good, just a few questions for clarification

terraform/stacks/agha_infra/main.tf
williamputraintan: Yes, this is the existence resource. perhaps we could remove this

terraform/stacks/agha_infra/main.tf
williamputraintan: With the current deployment, staging and results bucket have the `Bucket owner enforced` but the store bucket still have the `Object writer`. So to enable this at the store bucket, I enabled this here, but I added for all 3 buckets for consistency.

I think the default/recommendation is to make `bucket-owner-enforced` and it seems that not specifying this will just not touch what is currently deployed.

terraform/stacks/agha_infra/main.tf
williamputraintan: Ahh right could do! Changing it, thanks.

terraform/stacks/agha_infra/main.tf
reisingerf: OK, thanks!
I'm happy to have it explicit if it enforces the bucket owner. 
I am just trying to remember why the store is set to object writer... especially on the store bucket I think we want to be the ones having the full ownership, no?

terraform/stacks/agha_infra/main.tf
reisingerf: Remove it and see how it goes. 
I think we had an SCP that prevents changing this (but it's under our control), for UoM accounts it won't be, so good to get things aligned where we can.

terraform/stacks/agha_infra/main.tf
williamputraintan: Yeah I think we should have the ownership. Andrew made a [comment](https://umccr.slack.com/archives/CCBADM4SW/p1698380172962759) on this last month.

reisingerf: Hm, there must have been a reason why we switched, but I can't remember now...
OK, then let's switch it back to owner enforced. We'll notice if something breaks... sooner or later... ;-)

PR #391, opened 2023-12-20
Update lab metadata tracking sheet script with 2024 support

reisingerf:


alexiswl:
Should we change the current year as well from 2022 to 2024?

scripts/umccr_pipeline/UMCCR_Library_Tracking_MetaData.js
```diff
@@ -2,7 +2,7 @@
 // Should be attached to the UMCCR_Library_Tracking_MetaData Google Spreadsheet
 
 var CURRENT_YEAR = '2022'
```
victorskl, line 4: `CURRENT_YEAR` has no longer usage within the script scope. We could remove it, perhaps.

victorskl (approved):
LGTM otherwise. ditto Alexis minor comment on CURRENT_YEAR.

reisingerf:
@victorskl do you remember why we added the `CURRENT_YEAR` ? 
I can't remember if it's used at all (does not seem so), so might as well remove it to avoid confusion.

victorskl:
I can't recall either, Flo. Just opening up the script in PyCharm or so, it shows that it is now dangling var.

reisingerf:
OK, removed it.
As there are no other issues, I'll merge next.

victorskl:
Related
- https://github.com/umccr/data-portal-apis/pull/653
- https://github.com/umccr/samplesheet-check-frontend/pull/40

alexiswl:
> @victorskl do you remember why we added the CURRENT_YEAR ?

I thought it might have something to do with which sheet is opened

PR #396, opened 2024-01-12
Added databricks secrets rotation

alexiswl:
Drops need for IAM role passthrough,

### DIFF
--- a/cdk/apps/databricks/.gitignore
+++ b/cdk/apps/databricks/.gitignore
@@ -1,0 +1,8 @@
+*.js
+!jest.config.js
+*.d.ts
+node_modules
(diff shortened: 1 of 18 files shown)


victorskl (approved):
LGTM! 

I'm taking that a short term cred strategy is better than a long term passthrough role. Will catch up with you separately in another time for that.

alexiswl:
The IAM role passthrough in Databricks isn't always compatible with Unity Catalog, hence going through this route. 

Thought two weeks between key rotations for an AWS user (who's permissions are pretty restricted as it is), was appropriate but happy to shorten / extend as appropriate.

PR #417, opened 2024-04-16
Fix Oncoanalyser bucket policy

reisingerf:
Add PutObjectTagging as allowed action.


victorskl:
Portal S3 event processing required `PutObjectTagging` permission on buckets that keeping track of. 

ditto Slack thread for detail report - https://umccr.slack.com/archives/CP356DDCH/p1713182039703639

victorskl (approved):
Thanks, Flo!

PR #424, opened 2024-05-17
Update RNAsum prod to 0.6.5

skanwal:
**First wait for https://github.com/umccr/infrastructure/pull/423 to be merged**
 
- linked to RNAsum r-package update [umccr/RNAsum@main (commits)](https://github.com/umccr/RNAsum/commits/main/)
- cwl-ica PR https://github.com/umccr/cwl-ica/pull/493
- infrastructure dev PR https://github.com/umccr/infrastructure/pull/419
- infrastructure stg PR https://github.com/umccr/infrastructure/pull/423

PR #425, opened 2024-05-17
Update Input Keys

skanwal:


pdiakumis (approved):
Ah I just understood how those need to match the inputs in https://github.com/umccr/cwl-ica/blob/main/tools/rnasum/0.6.5/rnasum__0.6.5.cwl
LGTM thanks 👍

victorskl:
ETA, I will review first thing on Monday.

reisingerf:
@skanwal @pdiakumis  Sorry, just for my understanding: 

The name change of the input parameters means a breaking change, but the use and purpose of the parameters seems to not have changed? 

From the outside I am seeing a breaking change and I don't understand the reason for it... (and that's for any consumer of the CWL workflow, not just the portal)

skanwal:
Yes, the purpose of the parameters is the same. At, CWL level, we have been keeping the parameter id and prefix same for the tools/pipeline we have developed. That was the main motivation behind updating the rnasum tool definition (so mostly for consistency). 

At portal level - I have had a chat with Victor and understand now why this can be a breaking change. The number of changes required to get this aligned didn't seem too bad. Happy to discuss this in a meeting.

reisingerf:
OK, I see... you want the same user experience for the standalone tool AND the wrapping CWL.

Then I can see two options:
- 1. honour the CWL interface and do the parameter mapping inside the workflow. Accept the slight mismatch of CWL parameters and avoid the breaking change.
- 2. align the CWL exposed parameters with the tool parameters for consistency. This will mean a breaking change though and the version number should change accordingly (major version change).

I'd be happy with either of those.

reisingerf:
Btw: if there are potential external users that would benefit from the harmonisation of the tool parameters with the CWL parameters, then I'd choose option 2.

skanwal:
I don't want to push for a major version release for portal for one specific workflow parameter update. 
Happy to revert these.

@pdiakumis, keen to know your thoughts too.

pdiakumis:
I have no strong feelings about this, I am not a CWL or infra dev so I'll leave those decisions to you folks. I'm happy to accomodate with any changes on the rnasum r-package side.

reisingerf:
Sorry guys! I didn't want to make this complicated!

1, this is not something Victor or I can help with.
2, a portal update will need to happen either way and the procedure for that (currently) is the same for a minor version as for the major version. So it does not matter to the portal. We need to validate the change anyway. 
3, someone changed the parameter names. I don't know why, but there may very well haven been a good/valid reason.
4, it's usually a good idea to keep different way of running the same tool consistent. It avoids confusion down the road.

So, my suggestion:
If the renaming of the parameters makes sense (they are clearer, more understandable, preferable in any way) and there are no other reasons to NOT upgrade the major version, I'd say go that route.

Otherwise, if there are no benefits in the name change or an update to the major version would cause any other trouble, then I'd update the tool itself and keep the consistency between tool and CWL and current naming.

victorskl:
Could we make a quick catch up call to draw the consensus and plan forward, pls. Another team work we need to do for best optimal route.

victorskl:
Team discussed and concluded to go with RNAsum-1.0.0 version bump. This is better for RNAsum future.

skanwal:
@victorskl - I have created a major release `1.0.0` for RNAsum cwl workflow and updated rnasum terraform stack to refer to it.

Related cwl-ica PR is https://github.com/umccr/cwl-ica/pull/495.

I have left prod version as is now and update that in the other opened PR https://github.com/umccr/infrastructure/pull/424.

victorskl (approved):
LGTM. Once merged, I will follow up with applying it to dev and stg.

On Portal API side, it also need to update accordingly with [issue 672](https://github.com/umccr/data-portal-apis/issues/672)

victorskl:
_(sorry for slow responses, balancing time with orcabus dev progresses)_

skanwal:
@victorskl - thanks. I'll tackle portal API side tomorrow.

victorskl:
Btw folks;

In future orcabus, this will handle better; such that workflow run input/output contract will be through "data schema" binding. That will be essence of tinkering at orchestration side.

Coincidentally, right in time about we are sprinting this story topic in this week.
- https://github.com/umccr/orcabus/pull/320
- https://github.com/umccr/orcabus/issues/257

Just for the context, FYI and for observation.

PR #433, opened 2024-06-06
modify status page callback and redirect url in dev env

raylrui:
So far, we just need have new domain applied in our dev env. 

So only apply the the new callback and redirect url 
`status_page_dev_domain = "status.${var.base_domain[terraform.workspace]}"` 
in the dev env first in our cognito service.

Link issue https://github.com/umccr/data-portal-status-page/issues/105

### DIFF
--- a/terraform/stacks/cognito_aai/app_status_page.tf
+++ b/terraform/stacks/cognito_aai/app_status_page.tf
@@ -6,26 +6,23 @@
 locals {
   status_page   = "data-portal-status-page"
 
-  status_page_domain = "status.data.${var.base_domain[terraform.workspace]}"
+  status_page_domain = "status.${var.base_domain[terraform.workspace]}"
 
   status_page_alias_domain = {
-    prod = "status.data.umccr.org"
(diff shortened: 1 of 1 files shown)


terraform/stacks/cognito_aai/app_status_page.tf
```diff
@@ -14,15 +14,18 @@ locals {
     stg  = ""
   }
 
+  # FIXME: remove this when status page and apis is ready in production, https://github.com/umccr/data-portal-status-page/issues/105
+  status_page_dev_domain = "status.${var.base_domain[terraform.workspace]}"
```
victorskl, line 18: Hey Ray; it is okay to update `status_page_domain` at line 9 as well as line 12. Instead of introducing this temporary var.

We have terraform workspace isolated. So, This is fine.

raylrui: sure,  will make the change now

victorskl (approved):
Neat! Happy for you to apply it to Dev account.

PR #440, opened 2024-06-18
feat(orcabus): add permissions for the filemanager ingest role to access pipeline buckets.

mmalenic:
Related to https://github.com/umccr/orcabus/pull/355 and https://github.com/umccr/orcabus/issues/354

### Changes
* Add access permissions for the orcabus file-manager ingest function to get pipeline cache objects.
    * (I think this is correct although untested).


reisingerf:
LGTM
At some point the FM will need more permissions, but let's start with these.
FYI @victorskl

reisingerf (approved):

PR #445, opened 2024-06-25
Move data archive stack under unimelb

reisingerf:
Consolidate TF stacks for Unimelb AWS accounts under a common folder "unimelb".


victorskl (approved):
Sound good.!

PR #448, opened 2024-06-27
Add CloudTrail to data account

reisingerf:
This adds a basic trail logging all S3 data events. 
We could refine in the future if/when needed.

### DIFF
--- a/terraform/stacks/unimelb/data_archive/main.tf
+++ b/terraform/stacks/unimelb/data_archive/main.tf
@@ -17,30 +17,43 @@ terraform {
 provider "aws" {
   region = local.region
 }
-provider "awscc" {
-  region = local.region
-}
-
 
 data "aws_caller_identity" "current" {}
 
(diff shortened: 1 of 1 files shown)


brainstorm:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html might be interesting to enable for audits.

victorskl (approved):

reisingerf:
@brainstorm are you looking at something in particular that's not covered by CloudTrail?
See: https://docs.aws.amazon.com/AmazonS3/latest/userguide/logging-with-S3.html

brainstorm:
> @brainstorm are you looking at something in particular that's not covered by CloudTrail? See: https://docs.aws.amazon.com/AmazonS3/latest/userguide/logging-with-S3.html

Ah, cool then, I thought one had to explicitly enable S3 server logs, bad memory!

PR #451, opened 2024-06-28
Feature/data access analyser event forward

reisingerf:
An addition to the data account Terraform stack

- refactor out a security template
- add EventBridge rule and related permissions / role to forward IAM AccessAnalyser events to the UMCCR org account default event bus (where they should be picked up by an existing rule and forwarded to our #security Slack channel


reisingerf:
@mmalenic @victorskl 
It's not straight forward to test that this actually works. During the security patrol, could you please keep an eye open for data account findings in the security channel and whether they are formatted meaningfully (e.g. whether the account ID reflects that the event comes from the data account).

victorskl (approved):
Nice.! will keep an eye on it.

PR #457, opened 2024-07-11
Data Archive TF: Updated BYOB bucket to allow HeadObject

victorskl:
* Upgraded dev_lo_access to dev_ro_access
* Added GetObjectAttributes for HeadObject API call from FileManager and Portal
  https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html

### DIFF
--- a/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
+++ b/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
@@ -255,7 +255,8 @@ data "aws_iam_policy_document" "production_data" {
     }
     actions = [
       "s3:ListBucket",
-      "s3:GetObject"
+      "s3:GetObject",
+      "s3:GetObjectAttributes"
     ]
     resources = [
       aws_s3_bucket.production_data.arn,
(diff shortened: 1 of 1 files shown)


terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
```diff
@@ -255,7 +255,8 @@ data "aws_iam_policy_document" "production_data" {
     }
     actions = [
       "s3:ListBucket",
-      "s3:GetObject"
+      "s3:GetObject",
```
reisingerf, line 258: Why `GetObject`?

victorskl: `GetObject` is the existing one, Flo. I am adding new `GetObjectAttributes` (i.e. for `HeadObject` purpose). This requires modifying/adding a comma here. (see left side in diff/compare view)

Having said, this is for FM Lambda role (similarly as Portal Lambda role below). Those service Lambda use case involves reading the content.

reisingerf (approved):

PR #459, opened 2024-07-12
ICA ENS TF: Added ICA ENS subscription SQS queue

victorskl:


reisingerf (approved):
LGTM

terraform/stacks/ica_ens/policies/sqs_ica_ens_policy.json
```diff
@@ -0,0 +1,13 @@
+{
+   "Version":"2012-10-17",
+   "Statement":[
+      {
+         "Effect":"Allow",
+         "Principal":{
+            "AWS":"arn:aws:iam::079623148045:root"
```
reisingerf, line 7: If we make the ILMN account id a parameter then we have a generic SQS allow policy. 
Not needed though.... there is only one policy here and that parameter does not change.

victorskl: 👌

PR #471, opened 2024-08-08
Remove legacy pipeline cache

reisingerf:
The pipeline cache bucket has been replaced with dedicated prod / stg / dev buckets and will be remove with this change.


victorskl (approved):
LGTM

PR #478, opened 2024-09-20
Networking TF: Reverted back public IP for prod only

victorskl:
* Still need it for pieriandx submission pipeline until OrcaBus O3 release


victorskl:
See https://umccr.slack.com/archives/C8CG6K76W/p1724159625121809

PR #479, opened 2024-10-04
feat: add object tagging permissions for orcabus file manager

mmalenic:
Related to https://github.com/umccr/orcabus/pull/585 and https://github.com/umccr/orcabus/pull/587

### Changes
* Adds `s3:GetObjectTagging` and `s3:PutObjectTagging` so that file manager can access the BYOB tagging: https://github.com/umccr/orcabus/pull/585.
* Adds actions that allow versioned object access: https://github.com/umccr/orcabus/pull/587.

Note that I've also included the `s3:GetObjectAttributes` and `s3:GetObjectVersionAttributes`. The file manager is not using these yet, but I think it will also be a good idea to ingest attributes from the metadata at some point, which will make use of these.

I'm pretty sure this will work (I've run `terraform validate` and `terraform plan`).

### DIFF
--- a/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
+++ b/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
@@ -142,7 +142,14 @@ data "aws_iam_policy_document" "production_data" {
     actions = [
       "s3:ListBucket",
       "s3:GetObject",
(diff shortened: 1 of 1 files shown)


victorskl (approved):
LGTM

reisingerf (approved):

PR #483, opened 2024-10-25
Portal TF: Added Batch EventBridge rule to exclude oncoanalyser events from orcabus

victorskl:


victorskl:
See https://umccr.slack.com/archives/C03ABJTSN7J/p1729746933675779

victorskl:
Related https://github.com/umccr/nextflow-stack/pull/115

terraform/stacks/umccr_data_portal/pipeline/oncoanalyser.tf
```diff
@@ -33,6 +33,10 @@ resource "aws_cloudwatch_event_rule" "oncoanalyser" {
       "jobQueue" : [
         { "prefix" : "arn:aws:batch:${local.region}:${local.account_id}:job-queue/nextflow-pipeline" }
       ],
+      "parameters": {
+        "orcabus": [ { "exists": false } ],
```
reisingerf, line 37: That would not conver cases where orcabus = false I guess, but since we don't have any of those any time soon this should work nicely!

reisingerf (approved):

terraform/stacks/umccr_data_portal/pipeline/oncoanalyser.tf
victorskl: Yup

PR #487, opened 2024-10-25
Feat: Add New Admin Cognito User Group

williamputraintan:
Create an `admin` group within the Cognito user pool. This allows some Authorisation power based on the user included in this group. The first use case will be in the OrcaBus.

### DIFF
--- a/terraform/stacks/cognito_aai/app_orcaui.tf
+++ b/terraform/stacks/cognito_aai/app_orcaui.tf
@@ -4,7 +4,7 @@
 #
 
 locals {
-  orcaui_page   = "orcaui"
+  orcaui_page = "orcaui"
 
   orcaui_page_domain = "orcaui.${var.base_domain[terraform.workspace]}"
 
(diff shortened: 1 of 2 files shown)


terraform/stacks/cognito_aai/app_orcaui.tf
```diff
@@ -44,7 +44,16 @@ resource "aws_cognito_user_pool_client" "orcaui_page_app_client" {
   allowed_oauth_flows_user_pool_client = true
   allowed_oauth_scopes                 = ["email", "openid", "profile", "aws.cognito.signin.user.admin"]
 
-  id_token_validity = 24
+
```
williamputraintan, line 47: This isn't relevant to this PR, but it complains if the validity units/default don't match up.

victorskl (approved):

reisingerf (approved):

victorskl:
Merge this too, Will. I will apply accordingly.

PR #490, opened 2024-10-29
Add analysis data archive bucket

reisingerf:
This will add an analysis data archive bucket in our UniMelb data account.

Note: 
- There is no write access configured (yet) and other access only for the FileManager ingest role.
- Versioning is enabled (since production data)
- All objects get immediately pushed to GLACIER_IR storage class

### DIFF
--- a/terraform/stacks/unimelb/data_archive/analysis_archive.tf
+++ b/terraform/stacks/unimelb/data_archive/analysis_archive.tf
@@ -1,0 +1,101 @@
+################################################################################
+# Local constants
+
+locals {
+  # The bucket holding all archived analysis data
+  analysis_archive_bucket_name = "archive-prod-analysis-${data.aws_caller_identity.current.account_id}-${data.aws_region.current.name}"
+}
+
+################################################################################
+# Buckets
+
+resource "aws_s3_bucket" "analysis_archive" {
+  bucket = local.analysis_archive_bucket_name
+
+  tags = merge(
+    local.default_tags,
+    {
+      "Name"=local.analysis_archive_bucket_name,
+    }
+  )
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "analysis_archive" {
+  bucket = aws_s3_bucket.analysis_archive.bucket
+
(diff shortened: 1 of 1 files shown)


terraform/stacks/unimelb/data_archive/analysis_archive.tf
```diff
@@ -0,0 +1,101 @@
+################################################################################
+# Local constants
+
+locals {
+  # The bucket holding all archived analysis data
+  analysis_archive_bucket_name = "archive-prod-analysis-${data.aws_caller_identity.current.account_id}-${data.aws_region.current.name}"
```
mmalenic, line 6: This bucket will need to be configured as an event source in orcabus too, if filemanager wants to ingest events?

victorskl (approved):
LGTM

terraform/stacks/unimelb/data_archive/analysis_archive.tf
reisingerf: Yes I forgot the event bus config... adding...

reisingerf: Added with https://github.com/umccr/infrastructure/pull/490/commits/d2ec42bb2ae8073eb2c0ba0da3a18ebf77a1d3b4

Note: this does not include the activation of the eventbridge integration at the bucket level (due to possible integration with ICA BYOB). Any subscriber (like the FileManager) will have to set up their own rules against the OrcaBus event bus.

mmalenic: Is the intent to have this contain just prod data? E.g. there won't be an `archive-dev-analysis-*` bucket?

reisingerf: Yes, we don't have plans to archive any `dev` data. That's a `prod` thing only.

victorskl:
Gentle ping @alexiswl - pls review

alexiswl (approved):

PR #493, opened 2024-11-02
Add object tagging permissions to ICA IAM user

reisingerf:
Since the FileManager attaches object tags it seems the default permissions granted to the ICA admin user for BYOB are not sufficient. 
This will add tagging related permissions to the IAM user attached policy to circumvent possible `AccessDenied` errors when trying to copy tagged objects while not having access to those tags.


alexiswl (approved):
Tried and tested!

PR #502, opened 2024-11-18
Update fastq data archive bucket settings

reisingerf:
This is to finalise the bucket configuration / setup of our archive buckets before we start the data migration / archival.

Note:
- archive tiers are split in two, for large and smaller files
- versioning is enabled. This may have significant cost implications as all object versions will be charged for a minimal period. It's the only way to guard against accidental deletion though (I think).
- version expiry is based on the minimal charged duration (GLACIER_IA), not on the duration of the bulk of the data (DEEP_ARCHIVE)
- the FileManager role is only given read access, the prod account only has list access and there is no dedicated write access yet


victorskl:
Related https://github.com/umccr/orcabus/pull/700

victorskl (approved):

PR #513, opened 2024-12-04
Update byob_ica_v2.tf

andrewpatto:
The account delegated to needs to be the dev account - not the data account


andrewpatto:
No rush on applying this - I am still actually hand editing that policy

reisingerf (approved):

PR #517, opened 2024-12-06
Portal TF: Route archive buckets notification to Portal SQS

victorskl:


victorskl:
@reisingerf Flo, as discussed.

PR #520, opened 2024-12-12
agha-infra: Deep Archive S3 for MM data

williamputraintan:
- Add S3 lifecycle policy for `MM_VCGS` and `MM_PATHWEST` to DEEP_ARCHIVE

cc: @reisingerf


andrewpatto (approved):
Looks good

PR #522, opened 2024-12-24
data_archive TF: Fixed lifecycle policy reflow ordering issue

victorskl:
* Wrapped with built-in sort function to keep it maintained the order
  of actions and resources. So that it won't get reordered by TF parser
  dynamically and retained reproducible TF apply across iteration.


victorskl:
@reisingerf Flo pls see if this works at your end, when you are back from holis. No urgency. Just the trick that used in portal TF stack lots.

victorskl:
Related https://umccr.slack.com/archives/C06SX3AK56F/p1734999591742269

reisingerf (approved):
Uh, nice! 
Happy to go with it.

PR #526, opened 2025-01-06
Networking TF: Added EC2 Instance Connect Endpoint (EICE)

victorskl:
* You may typically want to EICE enabled private network Security Group
  which can be looked up name `ssh_from_eice`. Add this SG to your VM
  instance created in private subnet. Then you no longer need public IP
  to connect to this instance. Just need to do like so.

  ```
  aws ec2-instance-connect ssh --connection-type eice --os-user [ubuntu|ec2-user] --instance-id [i-1234567replace]
  ```

  Or, punch local tunnel like so:

  ```
  aws ec2-instance-connect open-tunnel --instance-id [i-1234567replace] --local-port 2222
  ```

### DIFF
--- a/terraform/stacks/networking/main.tf
+++ b/terraform/stacks/networking/main.tf
@@ -93,7 +93,8 @@ module "main_vpc" {
 }
 
 module "main_vpc_endpoints" {
-  source = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"
+  source  = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"
+  version = "5.1.2"
 
   vpc_id             = module.main_vpc.vpc_id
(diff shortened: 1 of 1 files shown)


victorskl:
Patto, just FYI. I will replace the EICE you created in Dev account awhile with the resource from this stack. As EICE is limited to 1 resource per subnet per VPC.

victorskl:
See more https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2-instance-connect/index.html

andrewpatto:
Can we add an EICE for the AG account? I need to spin up a machine there to do some clinical NAGIM analysis - and would prefer it lives in private subnet

andrewpatto:
Or will this happen in every account already once deployed?

andrewpatto:
ignore me - is already deployed I can see..

Issue #530, opened 2025-01-23
Give steps_s3_copy_restore_share_access write access to pipeline cache bucket


alexiswl:
In order to restore from archive, we want to restore and copy from archive bucket to pipeline cache bucket. 

As discussed in the meeting, the restore+share role does not have access to write to the pipeline cache bucket?

In order to use the s3 copy steps to restore data from the archive bucket to the byob / cache bucket this role will need access to the unimelb data bucket.

PR #531, opened 2025-01-23
Added permissions for s3 copy to have permission to write to cache bucket

alexiswl:
Resolves https://github.com/umccr/infrastructure/issues/530

### DIFF
--- a/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
+++ b/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
@@ -23,6 +23,9 @@ locals {
   # The role that the orcabus file manager uses to ingest events.
   orcabus_file_manager_ingest_role = "orcabus-file-manager-ingest-role"
   orcabus_data_mover_role = "orcabus-data-mover-role"
+
+  # S3 Stops Copy Share role
+  steps_s3_copy_restore_share_role = "umccr-wehi-data-sharing-role"  # FIXME to be changed it to a more permanent data sharing role in future
 }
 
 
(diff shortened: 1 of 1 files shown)


andrewpatto:
Should we limit it to a "sharing" keyspace in the bucket..

alexiswl:
Sure, or a 'restore' keyspace?

andrewpatto:
Yes -name to be decided .. "restore" sounds good. @reisingerf ?

andrewpatto:
So the Get and List should be *, and the Put and Delete restricted I think.

andrewpatto:
Also, "s3:GetObjectVersion" seems to be missing from the lists.. if GetObjectVersionTagging is there - so should the ability to get the actual versioned objects too

reisingerf:
Yes, a restriction to a "restore" prefix would be good. Also happy with 'restore'.

As for the permissions, they are probably a bit too generous. I don't think we need to deal with version (of tags nor objects). Versioning at the moment is an emergency backup in case something went wrong and then it generally requires an admin to sort things out.

andrewpatto:
> As for the permissions, they are probably a bit too generous. I don't think we need to deal with version (of tags nor objects). Versioning at the moment is an emergency backup in case something went wrong and then it generally requires an admin to sort things out.

I think we were thinking that when it comes to S3 permissions - we can conceive of no situation where we want to give someone access to objects and tags - but not also give access to versioned tags and objects and attributes were they in use. So totally get that they

alexiswl:
Updated, split read / write statements and added restore prefix to write roels

andrewpatto (approved):

reisingerf (approved):
LGTM

alexiswl:
Are you able to apply this @reisingerf, I don't have permissions to the UoM buckets

PR #535, opened 2025-01-30
Update fastq archive bucket policy

reisingerf:
Add 'GetObjectTagging' permissions to prod cross account statement to allow FileManager metadata sync by admins

### DIFF
--- a/terraform/stacks/unimelb/data_archive/fastq_archive.tf
+++ b/terraform/stacks/unimelb/data_archive/fastq_archive.tf
@@ -174,7 +174,8 @@ data "aws_iam_policy_document" "fastq_archive" {
       identifiers = ["472057503814"]
     }
     actions = sort([
-      "s3:List*"
+      "s3:List*",
+      "s3:GetObjectTagging"
     ])
     resources = sort([
       aws_s3_bucket.fastq_archive.arn,


terraform/stacks/unimelb/data_archive/fastq_archive.tf
```diff
@@ -174,7 +174,8 @@ data "aws_iam_policy_document" "fastq_archive" {
       identifiers = ["472057503814"]
     }
     actions = sort([
-      "s3:List*"
+      "s3:List*",
+      "s3:GetObjectTagging"
```
mmalenic, line 178: Not strictly necessary for my use-case, but could be good to match the analysis bucket with other `GetObject*` access? https://github.com/umccr/infrastructure/blob/d887aa134c3ad817fd1d68c26a5340d63b4c9d95/terraform/stacks/unimelb/data_archive/analysis_archive.tf#L84-L86

mmalenic (approved):
Thanks! LGTM

reisingerf:
Sorry, got a bit distracted. This is deployed now.

terraform/stacks/unimelb/data_archive/fastq_archive.tf
reisingerf: The idea is to not allow access to the actual data. It's supposed to be an archive after all. 

We may make some exceptions, but at this point I'd like to keep it at the strictly necessary permissions...

PR #537, opened 2025-02-12
Update the lifecycle configuration of the prod BYOB bucket

reisingerf:
Add rules for production cache and restored prefixes to expire data.

The `cache` prefix should be emptied by the workflows themselves. The lifecycle rule is just a backup in case that does not happen.

The restored prefix has multiple rules with different retention times for predefined prefixes. The `restored` root prefix should serve as a fallback for anything outside these configured prefixes and should set the baseline for other configurations. 

This will make use of lifecycle overloading/overlapping and depends on AWS evaluates overlapping/conflicting rules (see [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-conflicts.html)).
Noteworthy: "S3 Intelligent-Tiering is favored by S3 Lifecycle over any storage class". This could cause some additional / unexpected cost depending on when objects are accesses / transitioned.

### DIFF
--- a/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
+++ b/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf
@@ -22,6 +22,7 @@ locals {
   icav2_staging_project_name    = "staging"
   icav2_development_project_name = "development"
 
+  temp_data_prefix = "cache/"
   restored_data_prefix = "restored/"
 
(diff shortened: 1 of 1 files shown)


victorskl (approved):
LGTM

reisingerf:
@alexiswl FYI: I'll merge and deploy those rules as they should have no significant effect on production workflows
If there are any concerns or issues I'm happy to roll this back.

PR #538, opened 2025-02-17
feat(tf-unimelb): add filemanager presign user permissions

mmalenic:
### Changes
* Adds the orcabus filemanager presign user permissions to data buckets.
* Related PR: https://github.com/umccr/orcabus/pull/862

### DIFF
--- a/terraform/stacks/unimelb/data_archive/analysis_archive.tf
+++ b/terraform/stacks/unimelb/data_archive/analysis_archive.tf
@@ -116,6 +116,23 @@ data "aws_iam_policy_document" "analysis_archive" {
     ])
   }
 
+  # Allow the presign user to presign URLs from this bucket.
+  statement {
+    sid = "orcabus_file_manager_presign_access"
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${local.account_id_prod}:user/${local.orcabus_file_manager_presign_user}"]
+    }
+    actions = sort([
+      "s3:ListBucket",
+      "s3:GetObject"
+    ])
+    resources = sort([
+      aws_s3_bucket.analysis_archive.arn,
+      "${aws_s3_bucket.analysis_archive.arn}/*",
+    ])
+  }
+
   # Allow the data mover access to copy to this bucket.
   statement {
     sid = "orcabus_data_mover_access"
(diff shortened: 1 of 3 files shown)


terraform/stacks/unimelb/data_archive/analysis_archive.tf
```diff
@@ -116,6 +116,23 @@ data "aws_iam_policy_document" "analysis_archive" {
...
+    actions = sort([
+      "s3:ListBucket",
+      "s3:GetObject"
+    ])
+    resources = sort([
+      aws_s3_bucket.analysis_archive.arn,
+      "${aws_s3_bucket.analysis_archive.arn}/*",
+    ])
+  }
+
```
mmalenic, line 135: It might not be the intent to allow presigning from the analysis archive bucket. I can remove this if so. Same goes for the fastq archive bucket.

mmalenic: Presigning would only work for classes that are instantly retrieval anyway, e.g. `Glacier IR`.

victorskl (approved):

terraform/stacks/unimelb/data_archive/analysis_archive.tf
reisingerf: Ideally sharing would only happen from the BYOB, but until we have sorted and automated that, there will be sharing requests for data in archive. 
However, we agreed restore FASTQs into BYOB if/when that's needed, so there is no need to include the fastq archive. (short lived purls are already possible for data that's not in archive tiers, right?)

mmalenic: > However, we agreed restore FASTQs into BYOB if/when that's needed, so there is no need to include the fastq archive. (short lived purls are already possible for data that's not in archive tiers, right?)

There's no way to signal to the API which credentials to sign with - as of now, for https://github.com/umccr/orcabus/pull/862, it's always using the user access key.

I could do something like: try access key credentials -> check if URL works, if not, try role credentials. I was thinking of adding something like this in anyway, because I don't want the filemanager to accidentally return unusable presigned URLs

mmalenic: > There's no way to signal to the API which credentials to sign with - as of now, for https://github.com/umccr/orcabus/pull/862, it's always using the user access key.

This now works with https://github.com/umccr/orcabus/pull/862/commits/8301a6d38e8101b64002a9fbff438fda547f79ca

reisingerf (approved):

reisingerf:
I have (partially) deployed those changes now. 
(the user does not exist in the production account yet, so the related statements cannot be applied)

PR #541, opened 2025-03-21
Give s3 steps copy role permissions to read from the byob dev bucket

alexiswl:
And give access to write to the restored data prefix


reisingerf (approved):

reisingerf:
Deployed.

PR #545, opened 2025-05-12
Remove deprecated portal access

reisingerf:
The legacy portal has been removed. 
The bucket policy has to be updated to no longer refer to the portal role.


victorskl (approved):
LGTM

PR #546, opened 2025-05-13
Add FileManager pre-sign access statement to fastq archive bucket

reisingerf:
Add a statement to the fastq archive bucket policy to allow the FileManager pre-sign user access (following the template from the BYOB cache bucket).

This should allow the OrcaUI to serve (unarchived) data from the fastq archive like MultiQC reports.
(See: https://umccr.slack.com/archives/C025TLC7D/p1747108301357219)

FYI @mmalenic

### DIFF
--- a/terraform/stacks/unimelb/data_archive/fastq_archive.tf
+++ b/terraform/stacks/unimelb/data_archive/fastq_archive.tf
@@ -123,6 +123,22 @@ data "aws_iam_policy_document" "fastq_archive" {
     ])
   }
 
+  statement {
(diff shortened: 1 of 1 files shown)


victorskl (approved):
LGTM

alexiswl:
What multiqc reports are going into the fastq archive bucket?

reisingerf:
I think that's just for the very old data... 
Nowadays they'd be part of the bcl convert interop qc workflow, so should end up in the analysis archive, right?

PR #547, opened 2025-05-26
feat: add ListBucketVersions for filemanager presign user

mmalenic:
### Changes
* Adds `s3:ListBucketVersions` so that the filemanager API can perform a `ListObjectVersions` operation.

Related to implementing https://github.com/OrcaBus/service-filemanager/issues/8


reisingerf (approved):

mmalenic:
Apologies, the ingest role needs to have this permission as well, because that's what's being used to perform the action. I've added it to the ingest role as well.

PR #551, opened 2025-08-06
Remove test data ica user

reisingerf:
As agreed, this will remove purpose specific ICA IAM user and refer control over the BYOBs to the default production ICA user instead.
Note: the research project specific ICA users will be removed in its own PR.


victorskl (approved):
LGTM

PR #554, opened 2025-08-20
Update unimelb test data bucket setup

reisingerf:
- fix wrong prefix in lifecycle rule
- add dedicatd IAM user for ICA


reisingerf:
Note: this does not (yet) include the access removal from the pipeline user.
If we want to isolate access to the bucket to the dedicated test data user, we still need to revoke that permission from the pipeline user.

victorskl (approved):
LGTM

PR #561, opened 2025-11-21
Restrict Bucket policy for refdata bucket

reisingerf:
This should address the Security Hub finding (S3.6) reported in the last DevOps meeting.


williamputraintan (approved):
LGTM

PR #565, opened 2025-11-26
Refine Security Slack Notification Template and Rules

williamputraintan:
- Filter IAM Access Analyzer finding to report `ACTIVE` finding only
- Forward GuardDuty finding to Slack if the severity is >=7 (High/Critical)

### DIFF
--- a/terraform/stacks/unimelb/data_archive/main.tf
+++ b/terraform/stacks/unimelb/data_archive/main.tf
@@ -16,6 +16,10 @@ terraform {
 # Configure the AWS Provider
 provider "aws" {
   region = local.region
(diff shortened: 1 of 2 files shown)


terraform/stacks/unimelb/data_archive/main.tf
```diff
@@ -16,27 +16,33 @@ terraform {
 # Configure the AWS Provider
 provider "aws" {
   region = local.region
+
+  default_tags {
```
reisingerf: Sweet, thanks!

reisingerf (approved):

PR #583, opened 2026-03-20
changed the schedule and bumped the version of the parquet cloudtrail…

andrewpatto:
… thing


reisingerf (approved):
LGTM
Same request as for management though, to separate out this extra setup from the org core stack.

victorskl (approved):
LGTM too!

PR #597, opened 2026-09-17
register per-app callback URLs for portal micro-frontends

raylrui:
## Summary

couple with PR for frontend infra PR https://github.com/umccr/orcahouse-ui/pull/4.

Adds the per-path callback and logout URLs the portal frontends need so a user signing in from
`/hub/` or `/orcahouse/` is returned to that app rather than the portal root. Additive: existing URLs
are preserved exactly, and no SSM parameter values change.

## Context

The UMCCR portal hosts several independently deployed frontends on one domain, each on its own URL
path prefix (`/`, `/v2/`, `/hub/`, `/orcahouse/`), all sharing the `orcaui-app-<workspace>` Cognito
app client so a single sign-in covers all of them.

Cognito matches `redirect_uri` against `callback_urls` **exactly** — no prefix or wildcard matching.
The client previously listed only the portal root, so an app that returns to
`https://portal.<stage>.umccr.org/hub/` would be rejected with `redirect_mismatch`. Each app now
derives its own redirect from its base path, so each of those URLs must be registered.

## Changes

**`terraform/stacks/cognito_aai/portal_apps.tf`** (new) — one `portal_app_paths` local listing the
path prefixes (`""`, `/v2/`, `/hub/`, `/orcahouse/`), shared by both app clients below. Mirrors
`pathPrefix` in `lib/portal/apps.ts` of `umccr/frontend-infrastructure-pipelines`; kept in a
dedicated file with a comment because it is the one thing both clients depend on.

**`app_orcaui.tf`** — `callback_urls` / `logout_urls` are now the product of the workspace's origins
and `portal_app_paths` (via `setproduct`), replacing the single hardcoded root URL. The `""` path
reproduces today's value exactly, so this is purely additive.

**`app_local.tf`** — the localhost dev client gets the same per-path callbacks against
`var.localhost_url`, so local development works at `http://localhost:3000/hub/` etc. The three local
SSM redirect parameters now reference `var.localhost_url` directly instead of `sort(callback_urls)[0]`
— same value, but no longer silently dependent on list ordering (the sort happened to return the right
element only while the list had one entry).

## Effect per workspace (verified)

| Workspace | `orcaui-app` callback + logout URLs |
|---|---|
| dev | `portal.dev.umccr.org` + `/v2/`, `/hub/`, `/orcahouse/` (1 → 4) |
| stg | same shape on `portal.stg.umccr.org` (1 → 4) |
| prod | both `portal.prod.umccr.org` and `portal.umccr.org`, × 4 paths (2 → 8) |

`/orcaui/oauth_redirect_*_stage` and the localhost redirect parameters are unchanged in every
workspace. `orca-ui` v1 is unaffected — it still sends and receives the portal root.


raylrui:
Test on dev about the auth ( logout and login ) process: https://portal.umccr.org/orcahouse/ .

Prod will deployed after merged. (approved)

reisingerf (approved):
LGTM

reisingerf:
FYI: @williamputraintan

